PR #18, opened 2014-01-28
alsa: Add support for mixer element indexes

mpecio:
Some drivers register multiple mixer elements with the same name.
Make the plugin aware of mixer element indexes to tell such mixer
elements apart.
## 

Tested on a soundcard with two "PCM" mixer elements.

Verified to work:
- switching between elements with the same name
- switching between elements with different names
- controlling elements with unique name :)
- saving and restoring configuration
- reading v3.4 configuration (defaults to the first mixer element matching old configuration)

### DIFF
--- a/src/alsa/alsa.c
+++ b/src/alsa/alsa.c
@@ -617,6 +617,7 @@ void alsa_open_mixer (void)
 
     snd_mixer_selem_id_alloca (& selem_id);
     snd_mixer_selem_id_set_name (selem_id, alsa_config_mixer_element);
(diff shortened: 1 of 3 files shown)


kaniini:
@jlindgren90 do you think this stuff is still relevant?

jlindgren90:
No idea, I've never had hardware that used identical names for different elements.  Seems potentially like something that could be fixed better at the driver level?  Not sure.

kaniini:
Me neither.  Looks like it needs to be rewritten anyway.

PR #24, opened 2014-05-26
AOSD: Some minor tweaks (permanent title, also when paused)

rapidrabbit:
Hi all,

I always missed the possibility to have a permanent on-screen information so I just added it.

Also I didn't see the advantage of not knowing which title was paused, so I changed that, too, showing a "Paused" hint and the current position.

Feel free to integrate it, if it seems ok, I'm using it and it for me it works :-)

(only directly after start the OSD isn't shown which I might address later).

So long


jlindgren90:
I don't think we'll be doing any further feature development on AOSD, unless it is first ported to no longer require GTK2 and X11.  Going forward, we should be relying on Qt to provide compatibility across X11+Wayland+Win32.

PR #38, opened 2014-12-08
Added window type configuration

tiagoshibata:
I usually keep Audacious as an "Above others" window on the edge of the screen, but disliked when it got iconified (eg. when using Ctrl + D, or minimize all windows). So I added a sub-menu to choose the window type when using the Winamp Classic Interface. Options are:
-Dock: Remove window from pager and taskbar and create it as a dock (above others and won't be ever iconified).
-Desktop: Remove from pager and taskbar. Create it as a desktop window (keep under other windows and never iconify).

Window Managers usually ignore changes to the window type after windows have been created, so a skins_restart () is required.

Bugs:
When using a Dock type window, the window doesn't receive keyboard focus. Tested under Openbox, I don't know if all WMs behave this way. I tried forcing focus using xdotool instead and it didn't work either. It would be great if someone had a solution.

### DIFF
--- a/src/skins/menus.cc
+++ b/src/skins/menus.cc
@@ -116,6 +116,12 @@ static const AudguiMenuItem playlist_items[] = {
     MenuCommand (N_("Refresh Playlist"), "view-refresh", GDK_KEY_F5, NO_MOD, action_playlist_refresh_list)
 };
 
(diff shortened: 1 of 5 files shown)


jlindgren90:
This is not right, Audacious is not a dock or a desktop and shouldn't claim to be.  Closing.

PR #43, opened 2016-02-20
Fix scrobbler2 network problem (issue #618)

tallicamike:
When the network check was performed, the server was returning a SSL Error "Illegal Parameter (47)", which seems to be an OpenSSL/curl bug [1]. To avoid this bug I enabled SSLv3 for the scrobbler
connection. Following this change the Last.fm server responded but it didn't know the "user.getRecommendedArtists" API method (it is not present in the current API [2]). I changed the tester method to "user.getInfo" and parsed the "username" received from the Last.fm server.

[1] https://bugs.launchpad.net/ubuntu/+source/curl/+bug/595415
[2] http://www.last.fm/api/intro


jlindgren90:
The getRecommendedArtists problem was fixed already.  OpenSSL or CURL bugs should be reported to those projects and not Audacious.  Closing.

PR #44, opened 2016-08-25
Update Catalan translation

rbuj:


radioactiveman:
See [merge request 24](https://github.com/audacious-media-player/audacious/pull/24).

PR #46, opened 2016-09-11
Add support for signed 24 bit packed little endian (S24_3LE).

cdealti:


jlindgren90:
Should be working in master now.

PR #47, opened 2016-09-24
configure: use correct logic for flac filewriter

gustavoz:
Otherwise it tries to use the flac filewriter when flac isn't available.

Signed-off-by: Gustavo Zacarias <gustavo@zacarias.com.ar>


jlindgren90:
Thanks!

PR #49, opened 2016-09-29
Plugin to switch ReplayGain Album/Track mode based on playlist order.

kevinlekiller:
Turns on ReplayGain Album mode if Shuffle is off, or Shuffle is on and Album Shuffle if on.

Turns off ReplayGain Album mode if Shuffle is on.

I don't mind changing the license to BSD or GPL v2 or LGPL v2, I wasn't sure what was the preferred license for audacious-plugins.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -624,6 +624,16 @@ if test "x$enable_mac_media_keys" != "xno"; then
     GENERAL_PLUGINS="$GENERAL_PLUGINS mac-media-keys"
 fi
 
+dnl rgswitch
+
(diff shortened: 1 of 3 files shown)


jlindgren90:
I think this would be better as an option in core, rather than a plugin.  That is, instead of the current "Album mode" checkbox, add a "Mode" dropdown with three options, "Album", "Track", or "Based on Shuffle".

kevinlekiller:
I think it's better as an option also, will see what I can do.

Thanks.

PR #50, opened 2016-09-29
cue: Parse more data from cue files.

kevinlekiller:
I mainly wanted the ReplayGain information, but added in the other stuff while at it.

### DIFF
--- a/src/cue/cue.cc
+++ b/src/cue/cue.cc
@@ -106,6 +106,7 @@ bool CueLoader::load (const char * cue_filename, VFSFile & file, String & title,
                 }
 
                 Rem * rem = cd_get_rem (cd);
+
                 if (rem)
                 {
                     const char * s;
@@ -167,6 +168,7 @@ bool CueLoader::load (const char * cue_filename, VFSFile & file, String & title,
             }
 
             Rem * rem = track_get_rem (cur);
+
             if (rem)
             {
                 const char * s;
--- a/src/cue/cue.cc
+++ b/src/cue/cue.cc
@@ -17,6 +17,7 @@
  * the use of this software.
  */
 
+#include <ctype.h>
 #include <string.h>
 #include <pthread.h>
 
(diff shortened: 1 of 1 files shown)


src/cue/cue.cc
```diff
@@ -101,16 +101,27 @@ bool CueLoader::load (const char * cue_filename, VFSFile & file, String & title,
...
 
                 if (rem)
                 {
                     const char * s;
-                    if ((s = rem_get (REM_DATE, rem)))
-                        base_tuple.set_int (Tuple::Year, str_to_int (s));
+                    if ((s = rem_get (REM_DATE, rem))) {
+                        // We should extract the year out of the string, nothing guarantees it will only be a year.
+                        base_tuple.set_int (Tuple::Year, str_to_int(s));
+                        base_tuple.set_str (Tuple::Date, s);
```
jlindgren90: How about setting Tuple::Year if the string is all digits (use g_ascii_isdigit() to check), otherwise setting Tuple::Date?

src/cue/cue.cc
```diff
@@ -101,16 +101,27 @@ bool CueLoader::load (const char * cue_filename, VFSFile & file, String & title,
...
+                    }
+                    if ((s = rem_get (REM_REPLAYGAIN_ALBUM_GAIN, rem)))
+                        base_tuple.set_gain (Tuple::AlbumGain, Tuple::GainDivisor, s);
+                    if ((s = rem_get (REM_REPLAYGAIN_ALBUM_PEAK, rem)))
+                        base_tuple.set_gain (Tuple::AlbumPeak, Tuple::PeakDivisor, s);
                 }
+
+                base_tuple.set_int (Tuple::NumSubtunes, tracks);
```
jlindgren90: This isn't what NumSubtunes is for.  Just omit this line.

src/cue/cue.cc
```diff
@@ -156,13 +167,15 @@ bool CueLoader::load (const char * cue_filename, VFSFile & file, String & title,
                     tuple.set_str (Tuple::Genre, s);
             }
 
-            Rem * rem = cd_get_rem (cd);
+            Rem * rem = track_get_rem (cur);
 
             if (rem)
             {
                 const char * s;
-                if ((s = rem_get (REM_DATE, rem)))
```
jlindgren90: I guess this was redundant since it's already set in the base_tuple.  Good catch.

src/cue/cue.cc
kevinlekiller: Alright, maybe a "Tracks" or "TotalTracks" can be added to Tuple?

I assumed that is what it was, will remove.

src/cue/cue.cc
kevinlekiller: Will do some testing with that and update, thanks.

jlindgren90:
I made is_year() a separate function for readability, and merged this.  Thanks!

PR #51, opened 2016-09-29
Plugin to display metadata (GTK).

kevinlekiller:
Shows metadata / tag information for the current song.

I don't mind changing the license to BSD or GPL v2 or LGPL v2, I wasn't sure what was the preferred license for audacious-plugins.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -74,7 +74,7 @@ TRANSPORT_PLUGINS="gio"
 if test "x$USE_GTK" = "xyes" ; then
     EFFECT_PLUGINS="$EFFECT_PLUGINS ladspa"
     GENERAL_PLUGINS="$GENERAL_PLUGINS alarm albumart lyricwiki playlist-manager search-tool statusicon"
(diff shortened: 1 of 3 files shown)


jlindgren90:
I like the idea of making this a plugin.  Then it's dockable, unlike the current "Song Info" dialog.  But it would be nice to see this include album art and tag editing capabilities, so that it could replace the current dialog completely.  In this case, I'd rather have one solution that includes 100% of the features we want rather than two similar solutions that each include a different 75%.

jlindgren90:
Or maybe just work on improving the Qt info widget/window, which is already a plugin?  The GTK UI works well today, but any major UI changes/improvements probably ought to be directed to the Qt codebase since that will be the default eventually.

kevinlekiller:
Alright, I might work on Qt stuff instead, I made this because of the lack of other Qt features (like customizing playlist columns for example).

kevinlekiller:
Forgot to close this.

PR #52, opened 2017-03-25
Speed and pitch effect: Add option to synchronize pitch with speed

daniel-araujo:
Added a checkbox in the settings window of the speed and pitch effect that when enabled makes the values for pitch and speed stay synchronized. That means if you were to change the value for speed, pitch would be set to the same value. The same happens vice-versa.

By default the option is disabled. Existing users that don't care about it will not be bothered.


jlindgren90:
This is a good idea, thanks.  I made a couple of tweaks:
1) Disable the sound-stretch algorithm entirely if speed is coupled to pitch
2) Disable the speed entry in the GUI if it's coupled to pitch (makes the code simpler)

PR #54, opened 2017-03-31
reorder systray icon playback controls

landroni:
This commit reorders the systray context-menu playback controls so that the Play/Pause function comes out on top of the least. 

The goals are: 
- make Play/Pause easily accessible to the user, especially useful when the systray is on top
- makes the menu consistent with the Playback top menu
- adopts the approach taken by other players like Sonata or Gnome MPlayer. 

I know this point has been discussed and dismissed in http://redmine.audacious-media-player.org/issues/705 , but I came up with this branch 'in case' there is a change of heart. 

Overall it seems to me that Play is most intuitively placed as top of that list in the context of a... context menu. But it's quite possible this is a 'just me' situation, and I _do_ have the systray on top...


landroni:
Thanks!

PR #56, opened 2017-04-04
remove hide/restore toggle in Qt systray

landroni:
The commit removes the Hide/Restore toggle in the Qt systray, making the Gtk and Qt menus consistent with each other.

### DIFF
--- a/src/statusicon-qt/statusicon.cc
+++ b/src/statusicon-qt/statusicon.cc
@@ -169,8 +169,3 @@ void StatusIcon::toggle_aud_ui ()
 {
     aud_ui_show (! aud_ui_is_shown ());
 }
-
-void StatusIcon::update_menu ()
-{
-    QList< QAction *> acts = menu->actions ();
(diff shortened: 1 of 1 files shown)


src/statusicon-qt/statusicon.cc
```diff
@@ -176,7 +173,4 @@ void StatusIcon::toggle_aud_ui ()
 void StatusIcon::update_menu ()
```
jlindgren90: You should get rid of update_menu(), it doesn't do anything now.

landroni:
See updated patch.

radioactiveman:
Better, but please remove the method and its usages completely. The current code does not even compile.

landroni:
Sorry! I usually check all my patches for compilation issues, but not this one -- I assumed the change was innocuous. 

Is it fine now?

radioactiveman:
Now it's fine, thanks.

PR #57, opened 2017-05-23
Correct silence-removal effect translation

foreigner-rub:
It`s so hard to find silence removal function, when it called "Noisereduction"


jlindgren90:
The translations are managed through Transifex:
http://www.transifex.net/projects/p/audacious/

It would be great if you could make the corrections there, and we will include them in the next release.  Thanks!

radioactiveman:
Closing, as the corrections have been made at Transifex. Thanks for the fixes.

PR #60, opened 2017-11-05
Unbundle Adplug

SoapGentoo:
* Derive from `CFileProvider` instead of modifying code.
* Copy over derived class `vfsistream`.
* Debian and co. all include Adplug as a package nowadays.


jlindgren90:
Looks good, thanks!

PR #63, opened 2018-01-17
Mac/CoreAudio improvements

RJVB:
This is a companion to https://github.com/audacious-media-player/audacious/pull/29

Like that PR this englobes a number of patches aimed at improving the Mac/Qt version of the player (plus the build system update). The main changes:

- fixes the CoreAudio plugin (it didn't play for me, and apparently not just for me)
- look, menu and systray tweaks; use of more appropriate fonts, alternatives for a few missing icons in the QtUI theme (text-labelled buttons), clickable playlist headers, rich-text support in the search widget.
- support for a simple form of bitperfect audio output over CoreAudio, where the current default output device is switched to the sample rate of the content being played (or an appropriate multiple that should reduce beating and aliasing).

I've considered adding a device selection UI like the one in the ALSA plugin, but dropped that idea for now as the Mac already has 3 easy mechanisms for audio device selection; users who have multiple devices are bound to be familiar with them.

It would be nice if the Sort menu actions were coupled to the playlist headers (and incidentally, got checkmarks to show the current sorting method).


RJVB:
A few screenshots:

Mac, "before" (but already with the "native-sized" icon toolbar):
audacious-mac-aqua-before
Note the use of the same BIG font everywhere and the missing icons in the plugin prefs dialog and the search widget.

Mac, after:
audacious-mac-aqua-after
The refresh button in the search widget now uses a Unicode symbol instead of an icon; the prefs. dialog uses traditional text labels instead of icons. Font sizes have been made widget-style dependent where most justified (cf. the comparison with iTunes below) and the search results are now presented in rich text comparable to how the GTk UI presents them.

X11, using the QtCurve theme:
audacious-x11-qtcurve-after
The aesthetics changes are much more subtle in this case so I didn't take a "before" screenshot.

For comparison, iTunes on Mac:
itunes-look

src/playlist-manager-qt/playlist-manager-qt.cc
```diff
@@ -324,7 +323,7 @@ static QToolButton * new_tool_button (const char * text, const char * icon)
     auto button = new QToolButton;
     button->setIcon (audqt::get_icon (icon));
     button->setText (audqt::translate_str (text));
-    button->setToolButtonStyle (Qt::ToolButtonTextBesideIcon);
+    button->setToolButtonStyle (Qt::ToolButtonFollowStyle);
```
jlindgren90: Only this isn't really a toolbar.  We don't want to follow the toolbar style here, we want to follow the button style.

src/qtui/menus.cc
```diff
@@ -82,7 +87,11 @@ QMenuBar * qtui_build_menubar (QWidget * parent)
 
     static const audqt::MenuItem playback_items[] = {
         audqt::MenuCommand ({N_("_Play"), "media-playback-start", "Ctrl+Return"}, aud_drct_play),
+#ifdef Q_OS_MACOS
+        audqt::MenuCommand ({N_("Paus_e"), "media-playback-pause", "Space"}, aud_drct_pause),
```
jlindgren90: No. Space doesn't work in all parts of the GUI (not in any text entry, for example), which is why we have the alternate shortcut here.

src/qtui/menus.cc
```diff
@@ -72,8 +72,13 @@ QMenuBar * qtui_build_menubar (QWidget * parent)
         audqt::MenuSep (),
         audqt::MenuCommand ({N_("Search _Library"), "edit-find", "Ctrl+Y"}, show_search_tool),
         audqt::MenuSep (),
+#ifdef Q_OS_MACOS
+        audqt::MenuCommand ({N_("About Audacious"), "help-about"}, aud_ui_show_about_window),
+        audqt::MenuCommand ({N_("Preferences..."), "preferences-system", "Ctrl+,"}, aud_ui_show_prefs_window),
```
jlindgren90: What's the point of changing the names here?

src/qtui/menus.cc
```diff
@@ -149,7 +158,14 @@ QMenuBar * qtui_build_menubar (QWidget * parent)
         audqt::MenuSep (),
         audqt::MenuCommand ({N_("_New"), "document-new", "Ctrl+T"}, pl_new),
         audqt::MenuCommand ({N_("Ren_ame ..."), "insert-text", "F2"}, pl_rename),
+#ifdef Q_OS_MACOS
+        // Command+W is also the system shortcut for closing windows, and thus too
+        // likely too much of a reflex that can lead to closing playlists by accident
+        // (or getting the "are you sure" dialog unintentionally too often).
+        audqt::MenuCommand ({N_("Remo_ve"), "edit-delete", "Ctrl+Shift+W"}, pl_close),
```
jlindgren90: If we make this change, it shouldn't be Mac-specific.  I'd like more input from others first, though.  The switch to Ctrl-T / Ctrl-W (imitating web browsers) was intentional back when we started the GTKUI.

src/qtui/playlist-qt.cc
```diff
@@ -60,6 +61,7 @@ PlaylistWidget::PlaylistWidget (QWidget * parent, Playlist playlist) :
 
     updateSettings ();
     header->updateColumns ();
+    setSortingEnabled (true);
```
jlindgren90: Sorting needs to be implemented in such a way that the new order makes it down to the underlying playlist.  I don't think that this approach does so?

src/skins-qt/skin.cc
```diff
@@ -472,6 +473,23 @@ QFont * qfont_from_string (const char * name)
...
+        {"Bold", QFont::Bold}
+    };
+    static QHash<const QString, QFont::Style> styleMap = {
+        {"Oblique", QFont::StyleOblique},
+        {"Italic", QFont::StyleItalic}
+    };
+    static QHash<const QString, QFont::Stretch> stretchMap = {
+        {"Condensed", QFont::Condensed},
+        {"Expanded", QFont::Expanded},
+    };
```
jlindgren90: I suspect that binary search (with a pre-sorted table created at compile-time) would be more efficient overall, taking into account CPU time + memory usage, than creating a hash at run-time.  tuple.cc has an example.

src/statusicon-qt/statusicon.cc
```diff
@@ -89,6 +89,9 @@ const audqt::MenuItem StatusIcon::items[] =
     audqt::MenuSep (),
     audqt::MenuCommand ({N_("_Open Files ..."), "document-open"}, StatusIcon::open_files),
     audqt::MenuCommand ({N_("Se_ttings ..."), "preferences-system"}, audqt::prefswin_show),
+#ifdef Q_OS_MACOS
+    audqt::MenuCommand ({N_("_Hide"), "hide-application"}, StatusIcon::toggle_aud_ui),
```
jlindgren90: Should be "Toggle" rather than "Hide", I would think ...

src/qtui/menus.cc
kaniini: They will show up in the appropriate places in the Mac menu that way.

RJVB:
Yes, something like that will do the trick. It's the cheap way which triggers Qt's internal text-based guessmatics. There's a better way which involves `QAction::setMenuRole()` which should be possible to integrate into the lowlevel function that does the actual QAction creation. It'd have the advantage that it's more future-proof (roles are ignored on platforms where they don't make sense, but where andwhat they mean can involve) and you don't need conditional code.

I just didn't find it worth the hassle (also given that Audacious has at least 2 ways to open the settings dialog already).

src/playlist-manager-qt/playlist-manager-qt.cc
RJVB: Where do you get the toolbar reference from? All this change does is to instruct the new tool*button* to follow the toolbutton style so that users aren't obliged to waste horizontal space on text labels if they don't want that.

src/qtui/menus.cc
RJVB: On Mac, Command+Comma opens the preferences dialog, so that wouldn't work here either. I'll see how easily I can get used to Command+Space .

src/qtui/playlist_model.cc
```diff
@@ -118,6 +118,7 @@ QVariant PlaylistModel::data (const QModelIndex &index, int role) const
     switch (role)
     {
     case Qt::DisplayRole:
+    case Qt::ToolTipRole:
```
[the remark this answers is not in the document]
RJVB: What's the difference?

src/qtui/status_bar.cc
```diff
@@ -92,8 +97,8 @@ void StatusBar::log_handler (audlog::Level level, const char * file, int line,
 {
     tiny_lock (& message_lock);
 
-    /* do not replace a message of same or higher priority */
-    if (level <= current_message_level)
+    /* do not replace a message of higher priority, or non-critical same-level message */
+    if (level <= current_message_level && level < audlog::Warning)
```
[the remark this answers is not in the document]
RJVB: Not hiding any warning and higher messages in the statusbar.

OTOH it's a choice to show only the first or the last of a series of such messages generated in a time interval, and there's no general rule that says which is better AFAIK. But OTOH if you show only the first the user doesn't even know if other messages are being generated, and even if you cannot read all messages that fact in itself is an indication that somethings is wrong (a fact you'd miss if only the 1st message is shown).

src/skins-qt/skin.cc
RJVB: Is that really a concern here, for a 2-element table that's created at startup?

src/statusicon-qt/statusicon.cc
RJVB: No, "Hide" is the proper platform terminology, but there's actually no need for an extra platform-specific item here. There's already a "Hide Audacious" item with shortcut in the Application menu (provided by the OS).

src/qtui/status_bar.cc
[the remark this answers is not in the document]
jlindgren90: I chose to display the first message rather than the last, because subsequent messages are often either giving more detail about the error (useless if you can't see the first line) or listing additional errors which are side effects of the first one.  It's not perfect, and I agree that there's no general rule to handle all cases perfectly.

Given the limited space in the status bar, some messages *will* be hidden at one point or another.  It would be nice to give an indication of this.  Maybe a suffix "(+ 2 more messages)" that you can click on to open the Log Inspector and see the hidden messages.  Or just make the whole status bar clickable when a warning/error message is displayed.

src/qtui/playlist_model.cc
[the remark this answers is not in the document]
jlindgren90: See the screenshot on the audacious-media-player.org home page for an example of what the info popup looks like.

src/qtui/playlist-qt.cc
jlindgren90: I think that setSortingEnabled() is meant to change only the view and not the underlying model, which is not what we want here.  What we'll probably need to do is intercept the header clicks, map those to the matching Playlist::SortType value, and call into libaudcore (Playlist::sort_by_tuple) to do the actual sorting.

src/skins-qt/skin.cc
jlindgren90: Probably not.  Your commit message says "More efficient lookups", so I was taking that at face value and wondering if it really is more efficient.  I would rather see a pull request to implement an actual font chooser in libaudqt, rather than making this stopgap code more efficient.

RJVB:
>I think that setSortingEnabled() is meant to change only the view and not the underlying model, which is not what we want here. 

I must admit that I actually do not have much of an idea, but I would be surprised if Qt doesn't provide at least the hooks to keep contents and their display settings synchronised.

I just don't have much time to dig around in such matters these days, but will keep it in mind.

RJVB:
>Probably not.  Your commit message says "More efficient lookups", so I was taking that at face value and wondering if it really is more efficient.

It is more efficient than a traditional implementation where you do series of string comparisons - a bit at runtime and a bit (more) when coding.

>I would rather see a pull request to implement an actual font chooser in libaudqt, rather than making this stopgap code more efficient.

In the end this isn't really different from the typical GUI design approach that uses an interface editor. I haven't gone the font chooser path because I'm not convinced that it's required or would be more fun to code than what I've done now. Besides, you'd still want the default to look good and in my book that would boil down to keeping my current code.

kaniini:
What's the current situation with these patches, incidentally?  Are there any still worth looking at?

RJVB:
Hi,

On the one hand I'd love to say "I hope so" and "feel free to pick what you want". On the other hand I have been doing completely different things, getting more afraid with every passing month that I'd have to spend lots of time to refactor my patches (but without ever getting around to checking that).

The naked truth behind that is that I simply don't use Audacious as the free Foobar2000 player does exactly what I need and runs just fine on cheap hardware that comes with the required OS preinstalled...

PR #64, opened 2018-04-12
Add "Description" to "Song Info" window for OGGV and FLAC files.

DavidGriffith:
There's an ongoing problem with sort menus not working right, but I feel it's time to get this pulled in now and allow others to address those sort menus.


DavidGriffith:
I don't know what the problem is with Travis.  It compiles fine for me on Debian 9.

kaniini:
I merged parts of this (not the sorting parts).

PR #65, opened 2018-05-22
Make mpris trackid type a dbus object path

acrisci:
Mpris Track_Id type should be a dbus object path, not a string.

https://specifications.freedesktop.org/mpris-spec/latest/Player_Interface.html#Simple-Type:Track_Id


acrisci:
Fixes [this issue](https://github.com/acrisci/playerctl/issues/65) on playerctl.

jlindgren90:
Looks reasonable. Thanks!

PR #66, opened 2018-09-21
hotkey: Seek using step size from UI

adalessandro:
The hotkey plugin has a hardcoded 5 seconds seek step. This PR introduces a simple modification to read and use the step size setting value from the UI.

TODO: The language po/ files need to be updated. If this modification is approved I'll squash the language update into this same commit.


jlindgren90:
Sorry for the delay in getting back to you about this.  There was a bit more technical debt that needed to be addressed to really get this working right.  Various plugins (like gtkui and qtui) each had their own "step_size" setting.  I see you borrowed the gtkui setting here, but accessing one plugin's setting from another plugin isn't good (for a couple of reasons).  The step size for seeking and the volume step both ought to be global settings so that they are consistent throughout Audacious.

I added the global settings and changed over a bunch of other plugins that had hard-coded step sizes to use the new settings.  Let me know if anything still isn't working for you.

.pot files will be updated later in the release cycle so that new translations can be submitted (via Transifex).  We don't generally use GitHub for those.

PR #67, opened 2018-10-15
song_change plugin for Win32

domenn:
Song change now works under Windows. I tested, it works. Builds on both Linux and Windows.

### DIFF
--- a/src/songchange/songchange_crossplatform.h
+++ b/src/songchange/songchange_crossplatform.h
@@ -1,0 +1,136 @@
+//
+// Created by domen on 15. 10. 2018.
+//
+
+#ifndef AUDACIOUS1_SONGCHANGE_CROSSPLATFORM_H
+#define AUDACIOUS1_SONGCHANGE_CROSSPLATFORM_H
+
+#include <libaudcore/runtime.h>
+
+#ifdef _WIN32
+
+#include <string>
+#include <windows.h>
+
+std::string utf16_to_utf8(const std::wstring &utf16)
+{
+    // Special case of empty input string
+    if (utf16.empty())
+    {
+        // Return empty string
+        return {};
+    }
+
+
+    // "Code page" value used with WideCharToMultiByte() for UTF-8 conversion
+    const UINT codePageUtf8 = CP_UTF8;
+
+    // Safely fails if an invalid UTF-16 character is encountered
+    const DWORD flags = WC_ERR_INVALID_CHARS;
+
+    // Get the length, in chars, of the resulting UTF-8 string
+    const int utf8Length = ::WideCharToMultiByte(
+            codePageUtf8,       // convert to UTF-8
+            flags,              // conversion flags
+            utf16.c_str(),  // source UTF-16 string
+            utf16.size(),  // length of source UTF-16 string, in WCHARs
+            nullptr,            // unused - no conversion required in this step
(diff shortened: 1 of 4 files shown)


src/songchange/songchange_crossplatform.h
```diff
@@ -0,0 +1,136 @@
...
+#define AUDACIOUS1_SONGCHANGE_CROSSPLATFORM_H
+
+#include <libaudcore/runtime.h>
+
+#ifdef _WIN32
+
+#include <string>
+#include <windows.h>
+
+std::string utf16_to_utf8(const std::wstring &utf16)
```
jlindgren90: It looks like utf16_to_utf8() isn't used anywhere.

src/songchange/songchange_crossplatform.h
```diff
@@ -0,0 +1,136 @@
...
+        return {};
+    }
+
+    auto returning = std::string(bytes);
+    delete [] bytes;
+    // Return resulting UTF-8 string
+    return returning;
+}
+
+std::wstring utf8_to_utf16(const std::string &utf8) {
```
jlindgren90: Please use g_utf8_to_utf16():
https://developer.gimp.org/api/2.0/glib/glib-Unicode-Manipulation.html#g-utf8-to-utf16

src/songchange/songchange_crossplatform.h
```diff
@@ -0,0 +1,136 @@
...
+    ZeroMemory(&si, sizeof(si));
+    si.cb = sizeof(si);
+    ZeroMemory(&pi, sizeof(pi));
+
+    unsigned long flags = CREATE_UNICODE_ENVIRONMENT |  CREATE_NO_WINDOW;
+    auto * windows_cmd = const_cast<wchar_t *>(fileToRun.c_str());
+    BOOL executed = ::CreateProcessW(NULL, windows_cmd, NULL, NULL, TRUE, flags, NULL, NULL, &si, &pi);
+
+    if (!executed) {
+        AUDINFO((std::string("SongChange cannot run the command: ") + cmd + "\n").c_str());
```
jlindgren90: AUDINFO is a printf()-like macro, so you can use `%s` to print `cmd` rather than string additions.

src/songchange/songchange_crossplatform.h
```diff
@@ -0,0 +1,136 @@
+//
+// Created by domen on 15. 10. 2018.
```
jlindgren90: New files need to have a license header.  The song change plugin is GPL3 according to formatter.cc, so you could just copy the header from that file:
https://github.com/audacious-media-player/audacious-plugins/blob/master/src/songchange/formatter.cc

domenn: Should I use current year in copyright text?
/*  Audacious
 *  Copyright (C) 2005-**2018**  Audacious team
 ...

domenn:
Thank you for the feedback.

jlindgren90:
Thanks, this looks good.  I will try it out later this week and merge the changes.

jlindgren90:
Works great!  Thank you!

domenn:
Thanks for accepting.
Let me just say that Audacious is a great music player and since I use both OSes, I was finding something simple and cross platform, but some missing functionalities on Windows were dealbreaker. This one and UTF-8 in audtool will make it a lot more useful.

PR #68, opened 2019-02-04
Add a libopenmpt-based tracker module plugin.

cspiegel:
This plugin should support all files that are suppored by the
modplug-based plugin. However, due to how widespread libmodplug is, its
plugin is not removed, but instead set at a slightly lower priority, so
that if both it and the libopenmpt-based plugin are installed,
libopenmpt will be preferred. This allows systems without libopenmpt to
still have module playback support.


jlindgren90:
Thank you for the pull request!  I will try to take a closer look later this week.

One initial question.  Would it make sense to add this plugin in two phases: make it an option (but lower priority than modplug) in the next release, and wait till the release after to bump up its priority?

cspiegel:
That's a really good point I didn't consider.  Although openmpt supports all formats that modplug supports, due to the nature of MOD files, they won't necessarily sound the same in all circumstances, and I can see people being surprised at any potential changes, even if openmpt's playback is more accurate.

I agree that the introduction of openmpt should probably have something of a probationary period, so at least nobody is completely blindsided.  I'll move modplug's priority back to default, and drop openmpt's.

jlindgren90:
Thanks for the updates, and sorry I've been tardy about reviewing this.  I plan to sit down and look it over tomorrow.

jlindgren90:
@cspiegel Thank you much for submitting this!  Your C++ is very clean and was a pleasure to read.  I've made some further updates in master to simplify things a bit and make better use of some Audacious-specific types/features, as (IMHO) is suitable for an in-tree plugin.  It seemed easier to go ahead and make those changes myself rather than requesting them in review here, so I hope that's all right with you.  If I broke something or if any of my changes are odious to you for some reason, please feel free to say so.

JJRcop:
For convenience, here's a diff of the master changes https://github.com/audacious-media-player/audacious-plugins/compare/8638d7352c32d42c15e776bde9dd1610d12274d3...8969e115a8af8c20c140f48fdccf0bd2168cd624

cspiegel:
@jlindgren90 No problem at all bringing the plugin more in-line with the existing style and conventions.  Thanks for merging!

PR #69, opened 2019-06-20
psf: Add ignore length option for endless looping

tatokis:
This PR adds an option in the psf plugin to ignore the length of the tracks, resulting in endless looping.
It has been tested with psf and psf2 files. Other plugins have similar options.

### DIFF
--- a/src/psf/plugin.cc
+++ b/src/psf/plugin.cc
@@ -30,7 +30,9 @@
 
 #include <libaudcore/i18n.h>
 #include <libaudcore/plugin.h>
+#include <libaudcore/preferences.h>
 #include <libaudcore/audstrings.h>
+#include <libaudcore/runtime.h>
 
 #include "ao.h"
 #include "corlett.h"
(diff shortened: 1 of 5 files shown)


src/psf/plugin.cc
```diff
@@ -141,6 +163,9 @@ bool PSFPlugin::read_tag(const char *filename, VFSFile &file, Tuple &tuple, Inde
     return true;
 }
 
+extern void setendless(int e);
+extern void setendless2(int e);
```
jlindgren90: These should be declared in peops/spu.h and peops2/spu.h.

jlindgren90:
@kaniini Does this look okay to you?

src/psf/plugin.cc
tatokis: setlength() and setlength2() aren't in spu.h. They are externed when needed, which is why I followed the same practice.

https://github.com/audacious-media-player/audacious-plugins/blob/8d053afa9ee8327583b8521a14db481e61d883d5/src/psf/eng_spx.cc#L54

https://github.com/audacious-media-player/audacious-plugins/blob/8d053afa9ee8327583b8521a14db481e61d883d5/src/psf/eng_psf.cc#L69

If you think that is the correct way to do it, I have no objection and I'll add them to spu.h

kaniini: yeah please put it in spu.h

kaniini:
it looks okay with the requested change added.

src/psf/plugin.cc
tatokis: The compiler can't find them when put in spu.h, and I'm not sure it'll be a good idea to include both spu.h files in plugin.cc.
Since it gets included directly, would eng_protos.h be more appropriate?

jlindgren90: I fixed up the existing extern declarations.  peops/spu.h and peops2/spu.h are now included from plugin.cc, so please put the declarations in those headers.

tatokis: Thank you. I fixed the declarations

PR #78, opened 2019-12-19
adplug: Added more OPL emulators

KyleJ61782:
Added the Nuked, Woody and Ken Silverman OPL emulators. Also updated the
preferences GUI to handle the special case Nuked emulator that only
outputs in 16-bit stereo.


KyleJ61782:
Well, after looking at the CI and looking into the build failure, it seems that the Nuked and Woody OPL emulators only come in with adplug 2.3.0 and later. Since libadplug *is* a static lib, it would be possible to pull the github and build using that in the CI, but of course that's a different can of worms that may not be desirable.

mschwendt:
> Since libadplug is a static lib

It isn't a static lib. Multiple distribution even build with --disable-static as to build only the shared lib.

kjpowerworld:
Yeah, actually I just realized it's probably because I'm building on mingw-w64. It seems to me the proper way to handle what I'm looking to do is to build checks into the autoconf to check if various headers are available. I'm going to look at that now.

KyleJ61782:
Yeah, actually I just realized it's probably because I'm building on mingw-w64. It seems to me the proper way to handle what I'm looking to do is to build checks into the autoconf to check if various headers are available. I'm going to look at that now.

src/adplug/adplug-xmms.cc
```diff
@@ -291,14 +328,139 @@ const char * const AdPlugXMMS::defaults[] = {
...
-    WidgetCheck (N_("Duplicate mono output to two channels"),
-        WidgetBool (CFG_ID, "Stereo")),
+    WidgetCombo (N_("OPL Emulator:"),
+        WidgetInt (CFG_ID, "Emulator", emulator_changed),
+        {{plugin_combo}}),
+#ifdef USE_GTK
+    WidgetCustomGTK (create_optional_checks),
+#endif
+#ifdef USE_QT
+    WidgetCustomQt (create_optional_checks),
```
kaniini: Since these checks are harmless to display even with emulation engines that do not support them, please just declare the widgets in the toolkit-agnostic way.

src/adplug/adplug-xmms.cc
```diff
@@ -171,15 +192,30 @@ bool AdPlugXMMS::play (const char * filename, VFSFile & fd)
   dbg_printf ("open, ");
   open_audio (bit16 ? FORMAT_16 : FORMAT_8, freq, stereo ? 2 : 1);
 
-  CEmuopl opl (freq, bit16, stereo);
+  Copl *opl = nullptr;
```
kaniini: I would prefer to use a `SmartPtr` here.

src/adplug/adplug-xmms.cc
KyleJ61782: As in removing the ifdefs? That's doable. The optional checkboxes have to be done on a per-toolkit basis so that they can be hidden or shown depending on which OPL renderer is used as the Nuked renderer doesn't support 8 bit or non-stereo output.

src/adplug/adplug-xmms.cc
KyleJ61782: That's easy enough. When I get opportunity I'll do that.

src/adplug/adplug-xmms.cc
KyleJ61782: After rereading your comment, I now understand what you're getting at. While, yes, leaving the checkboxes doesn't hurt anything, I'm a firm proponent that a GUI should reflect underlying capabilities instead of misrepresenting them. So I'd either say that the checkboxes should be enabled or disabled according to the capabilities of the underlying plugin, or the checkboxes should just be removed as it's rather meaningless to support 8-bit and non-stereo rendering.

kaniini: Lets just remove the checkboxes for now.  If they were critical, I would suggest that we could extend the abstraction layer, but I don't think it is worth it for 8-bit audio output ;)

mschwendt: > as it's rather meaningless to support 8-bit and non-stereo rendering.

For Mono/Stereo, it depends on various factors: (1) how it's done by the internal audio mixer, (2) whether the emulated audio hardware did have two output channels, (3) whether original users connected the audio hardware to basic speakers built into monitors, and (4) whether many of the original users deliberately decided to ignore hardware output restrictions and used audio devices that mixed everything to Mono output for an overall better listening experience. For example, if using headphones, to hear the drums/percussions channel only on the left and other instruments on the right side wasn't anything like pretty, so many people preferred Mono output.

Typically, 8-bit output in emulators is available as an option for users of 8-bit audio hardware, which should not be an issue anymore, and for purists, who claim that they hear the differences in audio converted to 16-bit or higher precision.

kaniini: In reality, the 8-bit output is pointless anyway because we scale everything up to 16-bit minimum.

kaniini: Dropped the 8-bit output support as it is entirely pointless, as the core will change it to floating point anyway.

src/adplug/adplug-xmms.cc
KyleJ61782: Changed in commit 3d8642b.

src/adplug/adplug-xmms.cc
KyleJ61782: Removed checkboxes in 8bedf01.

configure.ac
```diff
@@ -189,6 +189,16 @@ ENABLE_PLUGIN_WITH_DEP(adplug,
     ADLIB,
     adplug)
 
+if test $have_adplug = yes ; then
```
kaniini: It would be nice to update the Meson side too.

src/adplug/Makefile
```diff
@@ -9,7 +9,9 @@ plugindir := ${plugindir}/${INPUT_PLUGIN_DIR}
...
 LD = ${CXX}
 CFLAGS += ${PLUGIN_CFLAGS}
-# FIXME: Turning off warnings for now; this code is awful
-CXXFLAGS += ${PLUGIN_CFLAGS} -Wno-sign-compare -Wno-shift-negative-value
-CPPFLAGS += ${PLUGIN_CPPFLAGS} ${ADLIB_CFLAGS} -I../..
-LIBS += ${ADLIB_LIBS}
+CXXFLAGS += ${PLUGIN_CFLAGS}
+CPPFLAGS += ${PLUGIN_CPPFLAGS} ${GTK_CFLAGS} ${QT_CFLAGS} ${ADLIB_CFLAGS} -I../..
+LIBS += ${QT_LIBS} ${GTK_LIBS} ${ADLIB_LIBS}
+ifeq ($(USE_QT),yes)
```
kaniini: this depends on GNU make, i think we should avoid that

configure.ac
KyleJ61782: In progress... #79 is a component of this. It appears meson builds were broken on mingw32 and potentially other platforms.

KyleJ61782: Fixed in efa1585.

KyleJ61782: Fixed in 6258fd6.

src/adplug/Makefile
KyleJ61782: Fixed in e466681.

kaniini:
Looks good to me, needs a final rebase though.

KyleJ61782:
I've fixed the conflicts. Should be good to go.

PR #80, opened 2020-01-05
Add HiDPI support to QT album art plugin

DmitriK:
Made album art plugin be aware of device pixel ratio for drawing. Note that this can result in pretty small pictures if the album art is smaller than the pixbuf size. It might be worth scaling up pictures by the device pixel ratio for smaller images.


kaniini:
If you add the scaling for small images, I'll merge this.

PR #83, opened 2020-02-23
Improved calculation of bitrate, vbr field of mpg123 is unreliable.

jeroenst:
After some more investigation I found that when not doing a full scan, the length is not accurate causing a lower bitrate. When calculating bitrate from  length, the value was not rounded but truncated, causing to drop 1kbps.

Now the bitrate is rounded (with one digit) which gives a correct bitrate for VBR and CBR.

Also I moved the song info applying to the block when not streaming, otherwise the song info is updated over and over when playing the song (maybe I had to make a separate commit for this?).

Please check, test and let me know if it's okay like this.

I tested with my cbr files and the send vbr file and all work fine.

### DIFF
--- a/src/mpg123/mpg123.cc
+++ b/src/mpg123/mpg123.cc
@@ -247,26 +247,25 @@ static bool read_mpg123_info(const char * filename, VFSFile & file,
     if (!s.valid())
         return false;
 
-    tuple.set_int(Tuple::Bitrate, s.info.bitrate);
-    tuple.set_str(Tuple::Codec, make_format_string(&s.info));
-
-    const char * chan_str = (s.channels == 2)
-                                ? _("Stereo")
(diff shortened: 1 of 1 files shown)


jlindgren90:
Rounding is a good idea.  We have some template functions (aud::rdiv and aud::rescale) to do that consistently.  So I switched to using those in b8e46d03894532c75e4b048823fb4a140126fa67.

I am not sure about the other changes here.  In particular, moving stuff inside `if(!stream)` seems like it will leave some fields blank for MP3 streams.  Is that intentional, or am I misunderstanding something?

jeroenst:
What I experience is that stream is only active when playing a file. Updating songinfo over and over while playing seems to me like unwanted behaviour.

But if playing a stream instead of a file I don't know how this behaves (yet).

PR #84, opened 2020-02-24
Showing CBR, VBR or ABR in songinfo

jeroenst:
This is a nice feature imo..

### DIFF
--- a/src/mpg123/mpg123.cc
+++ b/src/mpg123/mpg123.cc
@@ -256,6 +256,8 @@ static bool read_mpg123_info(const char * filename, VFSFile & file,
     tuple.set_str(Tuple::Quality,
                   str_printf("%s, %d Hz", chan_str, (int)s.rate));
 
+    int vbr = MPG123_CBR;
+
     if (!stream && s.rate > 0)
     {
         int64_t samples = mpg123_length(s.dec);
@@ -265,8 +267,16 @@ static bool read_mpg123_info(const char * filename, VFSFile & file,
         {
             tuple.set_int(Tuple::Length, length);
             tuple.set_int(Tuple::Bitrate, aud::rdiv<int64_t>(8 * size, length));
+            if (aud::rdiv<int64_t>(8 * size, length) !=  s.info.bitrate) vbr = MPG123_VBR;
         }
     }
+    
+    if (s.info.vbr != MPG123_CBR)
+        vbr = s.info.vbr;
+
+    tuple.set_str(Tuple::Quality,
+                  str_printf("%s, %d Hz, %s", chan_str, (int)s.rate, 
+                  vbr == MPG123_CBR ? "CBR" : vbr == MPG123_VBR ? "VBR" : "ABR"));
 
     return true;
 }


src/mpg123/mpg123.cc
```diff
@@ -265,8 +267,14 @@ static bool read_mpg123_info(const char * filename, VFSFile & file,
         {
             tuple.set_int(Tuple::Length, length);
             tuple.set_int(Tuple::Bitrate, aud::rdiv<int64_t>(8 * size, length));
+            if (aud::rdiv<int64_t>(8 * size, length) !=  s.info.bitrate) vbr = MPG123_VBR;
         }
     }
+    
+    if (s.info.vbr != MPG123_CBR) vbr = s.info.vbr;
```
kaniini: As previously discussed, MPG123 only accurately detects VBR scenario when Xing tag is present ahead of the first frame.  Accordingly I don't think this logic is robust.

kaniini: Additionally, the variable assignment should be on a new line.

jeroenst: That's why it is only used when not cbr. Otherwise the bitrate in info is compared to the calculated bitrate to determine vbr. (yes there is a minor chance that it will be incorrect but lt when xing header is not present). 

Variable assignment will I change as you wish.

kaniini: But my point is, MPG123 tends to report CBR mode for any file lacking a Xing tag ahead of the first frame.

jeroenst: Yes, that's correct, that's why I only use the MPG123 info.vbr when it is not CBR because only then we know it's valid. Otherwise I use the comparison between calculated bitrate and bitrate reported by MPG123 (if they are equal it's very likely CBR otherwise it is VBR, and I never seen ABR before so i think we can leave that behind.

No it's not 100% waterproof, but I think that's not possible unless the whole file is scanned (all frames same size = cbr, different frame sizes = vbr) which takes more time.

jeroenst: > Additionally, the variable assignment should be on a new line.

Done, I pushed a new commit.

src/mpg123/mpg123.cc
```diff
@@ -265,8 +267,16 @@ static bool read_mpg123_info(const char * filename, VFSFile & file,
         {
             tuple.set_int(Tuple::Length, length);
             tuple.set_int(Tuple::Bitrate, aud::rdiv<int64_t>(8 * size, length));
+            if (aud::rdiv<int64_t>(8 * size, length) !=  s.info.bitrate) vbr = MPG123_VBR;
```
jlindgren90: This is just guessing.  It's easily broken by adding a large ID3 tag -- then the computed average bitrate is thrown off and a CBR file is reported by VBR.  I don't think we should do this.

jlindgren90:
I think we should not add this feature.  I don't think it is possible to reliably distinguish between CBR and VBR without a full scan (hurting performance).

kaniini:
Agreed.  I don't think it's worth it.

jeroenst:
Is it an idea to show cbr or vbr when full scan is enabled?

src/mpg123/mpg123.cc
jeroenst: Isn't this also a problem that occurs when calculating the bitrate?

kaniini:
I don't think it is worth it, in fact I am more inclined to remove full scan mode.

jlindgren90:
The full scan option is needed for playing .cue+.mp3 files.  Without it, seeking is very inaccurate and you get very noticeable skips/repeats between tracks.

PR #86, opened 2020-04-13
Implement hotkeys plugin for qt interface

iDarkTemplar:
Ported from hotkeys plugin for gtk interface.
Mouse buttons support dropped.

Closes: https://redmine.audacious-media-player.org/issues/820


acinclude.m4
```diff
@@ -218,7 +218,7 @@ AC_ARG_ENABLE(qt,
 if test $USE_QT = yes ; then
     PKG_CHECK_MODULES([QTCORE], [Qt5Core >= 5.2])
     PKG_CHECK_VAR([QTBINPATH], [Qt5Core >= 5.2], [host_bins])
-    PKG_CHECK_MODULES([QT], [Qt5Core Qt5Gui Qt5Widgets >= 5.2])
```
kaniini, line 221: This breaks compilation with autoconf on non-X11 systems.

Please split it into it's own dependency as we do with QtMultimedia for the qtaudio plugin.

configure.ac
```diff
@@ -533,6 +533,13 @@ ENABLE_PLUGIN_WITH_TEST(qtglspectrum,
     auto,
     VISUALIZATION)
 
+ENABLE_PLUGIN_WITH_DEP(qthotkey,
```
kaniini, line 536: This is wrong because a `QTX11EXTRAS` dependency hasn't been set up in the autoconf script.

src/qthotkey/Makefile
```diff
@@ -0,0 +1,19 @@
...
+include ../../extra.mk
+
+plugindir := ${plugindir}/${GENERAL_PLUGIN_DIR}
+
+LD = ${CXX}
+CFLAGS += ${PLUGIN_CFLAGS}
+CPPFLAGS += ${PLUGIN_CPPFLAGS} ${QT_CFLAGS} -I../.. -I..
+LIBS += ${QT_LIBS} -lX11
+
+moc_%.cc: %.h
```
kaniini, line 15: Using MOC is not allowed in audacious, in most cases you don't actually need it and can achieve the same results with subclassing.

src/qthotkey/gui.cc
```diff
@@ -0,0 +1,329 @@
...
+			if (keysym == 0 || keysym == NoSymbol)
+			{
+				text = QString::fromLocal8Bit("#%1").arg(key);
+			}
+			else
+			{
+				text = QString::fromLocal8Bit(XKeysymToString(keysym));
+			}
+
+			for (int j = 0; j < sizeof(modifiers)/sizeof(modifiers[0]); ++j)
```
kaniini, line 113: Use `aud::n_elems(modifiers)` here.

src/qthotkey/gui.cc
```diff
@@ -0,0 +1,329 @@
...
+	delete combobox;
+	delete keytext;
+	delete button;
+}
+
+QPrefWidget::QPrefWidget(QWidget *parent)
+	: QWidget(parent),
+	main_widget_layout(new QVBoxLayout(this)),
+	information_pixmap(new QLabel(this)),
+	information_label(new QLabel(QString::fromLocal8Bit(_("Press a key combination inside a text field.")), this)),
```
kaniini, line 156: `QString::fromLocal8Bit()` is not necessary here, but you should use `audqt::translate_str()` to properly translate mnemonics to Qt style from GTK style, so that new strings are not added unintentionally.

src/qthotkey/gui.cc
```diff
@@ -0,0 +1,329 @@
...
+	[EVENT_JUMP_TO_FILE] = N_("Jump to file"),
+	[EVENT_TOGGLE_WIN] = N_("Toggle player window(s)"),
+	[EVENT_SHOW_AOSD] = N_("Show On-Screen-Display"),
+	[EVENT_TOGGLE_REPEAT] = N_("Toggle repeat"),
+	[EVENT_TOGGLE_SHUFFLE] = N_("Toggle shuffle"),
+	[EVENT_TOGGLE_STOP] = N_("Toggle stop after current"),
+	[EVENT_RAISE] = N_("Raise player window(s)")
+};
+
+class QLineKeyEdit: public QLineEdit
```
kaniini, line 77: Class names should not be prefixed with Q, as they are not part of Qt.

src/qthotkey/gui.cc
```diff
@@ -0,0 +1,329 @@
...
+#include <QtCore/QStringList>
+#include <QtGui/QKeyEvent>
+#include <QtGui/QMouseEvent>
+#include <QtWidgets/QApplication>
+#include <QtWidgets/QStyle>
+#include <QtX11Extras/QX11Info>
+
+#include <X11/XKBlib.h>
+
+static void destroy_callback();
```
kaniini, line 52: Consider whether this could be better expressed as a static class function of your preferences widget.

src/qthotkey/plugin.h
```diff
@@ -0,0 +1,50 @@
+#ifndef _PLUGIN_H_INCLUDED_
+#define _PLUGIN_H_INCLUDED_
+
+#include <QtCore/QList>
+
+typedef enum {
```
kaniini, line 6: typedef syntax here is redundant, use `enum Event { ... };`.

src/qthotkey/Makefile
kaniini: To follow up, I don't see any Q_OBJECT macros in your code, so the MOC is entirely unnecessary.

src/qthotkey/plugin.h
iDarkTemplar: Thanks, fixed it here.

src/qthotkey/gui.cc
iDarkTemplar: Thanks, fixed it here.

src/qthotkey/gui.cc
iDarkTemplar: Thanks, fixed.

src/qthotkey/gui.cc
iDarkTemplar: Replaced all _(...) calls with audqt::translate_str(...) calls. Kept N_(...) calls as they are.

There should be no new translatable strings except for one in the quoted code.
But this string change is intentional: since I dropped mouse buttons support, I also removed text mentioning this feature.

src/qthotkey/gui.cc
iDarkTemplar: Thanks, fixed.

src/qthotkey/Makefile
iDarkTemplar: There were a few Q_OBJECT macros, but I removed them, and it looks like they were no longer needed since plugin worked fine for me.

Thus I removed moc and qrc support from Makefile.

configure.ac
iDarkTemplar: You're right. Reworked it.

acinclude.m4
iDarkTemplar: You're right. Reworked it.

iDarkTemplar:
Thank you for quick review. I've added meson build support, and autotools build should be fixed too. Please let me know if change needs additional improvements.

src/qthotkey/plugin.h
```diff
@@ -0,0 +1,53 @@
+#ifndef _PLUGIN_H_INCLUDED_
+#define _PLUGIN_H_INCLUDED_
+
+#include <QtCore/QList>
+
+namespace GlobalHotkeys {
+
+enum EVENT {
```
kaniini, line 8: Use `enum Event` and camelcase enums like `Play`, `PrevTrack`, etc.  Then you can refer to your events as `Event::Play`, etc.

src/qthotkey/plugin.h
```diff
@@ -0,0 +1,53 @@
...
+	EVENT_TOGGLE_REPEAT,
+	EVENT_TOGGLE_SHUFFLE,
+	EVENT_TOGGLE_STOP,
+
+	EVENT_RAISE,
+
+	EVENT_MAX
+};
+
+struct HotkeyConfiguration {
```
kaniini, line 33: Most likely this could be done with a `QMap` or similar?

pevik:
@iDarkTemplar thanks for your contribution, I miss this plugin a lot :). @kaniini thanks for a quick review.

src/qthotkey/plugin.h
iDarkTemplar: It's a design approach I've taken from gtk global hotkeys plugin. The preferences widget doesn't have fixed slot for each action and instead it allows to add unlimited amount of hotkey combinations. It means, for one action it's possible to set a few different hotkey combinations.

Unless this design needs to be reworked and restricted to one hotkey combination per possible action, I'd prefer to not change it.

src/qthotkey/plugin.h
iDarkTemplar: I'll update it.

iDarkTemplar:
Check for x11 is present already:

https://github.com/audacious-media-player/audacious-plugins/blob/11b4c5cf9750bed9a6777ff7174938571a627644/src/qthotkey/plugin.cc#L404-L420

There's one more design issue to discuss: settings compatibility with hotkey plugin for gtk interface.

Currently I'm using same configuration values in this plugin. But since I've dropped support for mouse buttons in this plugin, if saved settings contain combinations with mouse buttons, it'll currently be silently converted to keyboard combination, with unknown result. I can make at least one of following changes:
1. different settings group can be used. This would disable migration of settings from gtk hotkeys plugin.
2. Check for hotkeys including mouse buttons can be implemented, and when such hotkeys are detected, they'd be silently dropped. If same settings group would be used as for gtk plugin, it might cause some issues if user used gtk version of plugin again.
3. Previous two approaches combined: use new settings group for this plugin, but if it's missing, try reading settings of gtk plugin for migration, skipping hotkeys using mouse buttons.
4. Don't change currently implemented behaviour.

Of course, some other ways to deal with this issue are possible.

kaniini:
You shouldn't be calling `audqt::init()` and `audqt::cleanup()` in your plugin, actually.  Those are explicitly for the UI plugins.

jlindgren90:
> You shouldn't be calling `audqt::init()` and `audqt::cleanup()` in your plugin, actually. Those are explicitly for the UI plugins.

It's fine to call those from general plugins. They are reference-counted. The notify plugin does the same thing.

iDarkTemplar:
> You shouldn't be calling `audqt::init()` and `audqt::cleanup()` in your plugin, actually. Those are explicitly for the UI plugins.

Initialization of this plugin doesn't properly work until QApplication is instantiated. So I either need to call `audqt::init()` and `audqt::cleanup()` or I need somehow to do a delayed initialization with no way to report whether this delayed initialization succeeded or failed back to audacious.

iDarkTemplar:
This is how `init()` looked without `audqt::init()`:

https://github.com/audacious-media-player/audacious-plugins/blob/5cba5ab13244f23e698d949b728803064f1f2094/src/qthotkey/plugin.cc#L117-L138

kaniini:
ah, I guess use `audqt::init()` then.

meson.build
```diff
@@ -42,9 +42,9 @@ xml_dep = dependency('libxml-2.0')
 if get_option('qt')
   qt_req = '>= 5.2'
   qt_dep = dependency('qt5', version: qt_req, required: true, modules: ['Core', 'Widgets', 'Gui'])
+  qtx11extras_dep = dependency('qt5', version: qt_req, modules: ['X11Extras'])
```
kaniini, line 45: this should have `required: false`.

src/qthotkey/gui.cc
```diff
@@ -0,0 +1,333 @@
...
+
+#include <libaudqt/libaudqt.h>
+
+#include <X11/XKBlib.h>
+
+namespace GlobalHotkeys {
+
+static PrefWidget *last_instance = nullptr;
+
+static const char *event_desc[static_cast<int>(Event::Max)] = {
```
kaniini, line 59: If you use a `QMap` here, you can avoid the static casts.

kaniini:
This seems like it is probably OK enough to go in at this point.

src/qthotkey/gui.cc
```diff
@@ -0,0 +1,334 @@
+/*
+ *  This file is part of audacious-hotkey plugin for audacious
+ *
+ *  Copyright (C) 2020 i.Dark_Templar <darktemplar@dark-templar-archives.net>
+ *  Copyright (c) 2007 - 2008  Sascha Hlusiak <contact@saschahlusiak.de>
+ *  Name: gui.c
+ *  Description: gui.c
+ *
```
pevik, line 8: nit: this is copy paste error (file is .cc), I'd delete these useless Name: and Description: lines.

pevik:
Thanks a lot for solving this :).

pevik:
Compiled and loaded manually. Works perfectly, it even remember my settings from GTK :). Great job, thanks!

emoxam:
No settings button in QT mode. audacious 4.4 on Kubuntu 22.04.4. Maybe it should appears as some menu item ?

radioactiveman:
Make sure the SVG module for Qt is installed. See also: https://audacious-media-player.org/problems

emoxam:
awesome! 
apt install libqt6svg6
helped!
thanks!

PR #88, opened 2020-05-24
Audio file format plugins: set channels information

iDarkTemplar:


iDarkTemplar:
This change depends on https://github.com/audacious-media-player/audacious/pull/38 to compile and work.

I've tested this change only on .mp3 files with 2 channels and .flac files with 1 channel and 2 channels, but most supported file types should display proper channels count as well, except for .cue files.

To display channels playlist has to be refreshed

src/amidiplug/amidi-plug.cc
```diff
@@ -150,15 +154,11 @@ bool AMIDIPlug::read_tag (const char * filename, VFSFile & file, Tuple & tuple,
 
     tuple.set_str (Tuple::Codec, "MIDI");
     tuple.set_int (Tuple::Length, mf.length / 1000);
+    tuple.set_int (Tuple::Channels, s_channels);
```
jlindgren90, line 157: read_tag() should not be using a static variable; it needs to be thread-safe.

src/ffaudio/ffaudio-core.cc
```diff
@@ -431,8 +431,13 @@ bool FFaudio::read_tag (const char * filename, VFSFile & file, Tuple & tuple, In
     if (! find_codec (ic.get (), & cinfo))
         return false;
 
+    ScopedContext context (cinfo);
+    if (LOG (avcodec_open2, context.ptr, cinfo.codec, nullptr) < 0)
```
jlindgren90, line 435: Not a fan of doing a stream-open just to get the channel count

src/amidiplug/amidi-plug.cc
iDarkTemplar: Would explicit call to function **backend_audio_info** be fine?

src/ffaudio/ffaudio-core.cc
iDarkTemplar: I don't see any other way to get information about channels here, but I'm not familiar with ffmpeg API. As alternative, changes for this file may be dropped.

src/amidiplug/amidi-plug.cc
jlindgren90: That seems fine.

src/ffaudio/ffaudio-core.cc
jlindgren90: Let's do that for now, just leave this one out.

iDarkTemplar:
Is anything still blocking this change?

PR #89, opened 2020-05-24
skins-qt: allow dropping files into playlist

iDarkTemplar:


src/skins-qt/playlist-widget.cc
```diff
@@ -738,6 +743,37 @@ bool PlaylistWidget::motion (QMouseEvent * event)
...
+    const QMimeData * mimedata = event->mimeData();
+
+    if (mimedata->hasUrls())
+    {
+        Index<PlaylistAddItem> files;
+
+        for (const auto &url: mimedata->urls())
+        {
+            if (url.isLocalFile())
+                files.append(String(url.toString().toLocal8Bit().data()));
```
jlindgren90, line 768: We should use toEncoded() here rather than toString().

jlindgren90:
Thanks for getting this working. I think a lot of Winamp skin users will appreciate having drag-and-drop working again. I've added a few more fixes in https://github.com/audacious-media-player/audacious-plugins/commit/c58c58d473005d94b7bda95765ae27e8969211c5.

PR #90, opened 2020-06-12
qtui: Implement wheelEvent for TimeSlider

SimonPilkington:
Allows seeking with the mouse wheel by placing the cursor over the
TimeSlider. This feature is already available in GTK mode.

### DIFF
--- a/src/qtui/time_slider.cc
+++ b/src/qtui/time_slider.cc
@@ -96,6 +96,19 @@ TimeSlider::TimeSlider(QWidget * parent)
     start_stop();
 }
 
+void TimeSlider::wheelEvent(QWheelEvent *event)
+{
+    m_scroll_delta += event->angleDelta().y();
+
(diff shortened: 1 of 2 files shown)


radioactiveman:
The seeking direction is different between Qt and GTK. Is this intended?
I am not sure though which one is more intuitive.

SimonPilkington:
Yes, you're right. It wasn't intended. This implementation is consistent with the volume slider and the equalizer (wheel up increases the value, wheel down decreases). The GTK implementation behaves like horizontal scrolling, which I think is bizarre.

radioactiveman:
Alright, this behavior sounds ok for me. Thanks for your contribution.

SimonPilkington:
You're welcome. To make GTK behave consistently making the values on line https://github.com/audacious-media-player/audacious-plugins/blob/master/src/gtkui/ui_gtk.cc#L411 negative should do.

PR #91, opened 2020-06-14
Directory listing plugin

mojyack:
A simple plugin that displays the directory structure. Clicking on a folder creates a playlist containing the music files under it.

### DIFF
--- a/src/library-view-qt/library-view-qt.cc
+++ b/src/library-view-qt/library-view-qt.cc
@@ -1,0 +1,83 @@
+#include <glib.h>
+
+#include <libaudcore/audstrings.h>
+#include <libaudcore/runtime.h>
+
+#include "library-view-qt.h"
+#include "tree-view.h"
+
+static QPointer<LibraryTreeView> tree_widget;
+EXPORT LibraryViewQt aud_plugin_instance;
+
+const char* const LibraryViewQt::defaults[] = {
+    "library_path",   DEFAULT_LIBRARY_PATH,
+    "playlist_title", DEFAULT_PLAYLIST_TITLE,
+    "playing_title", DEFAULT_PLAYING_TITLE,
+	"auto_send", "TRUE",
+    nullptr};
+
+const PreferencesWidget LibraryViewQt::widgets[] = {
+    WidgetLabel(N_("<b>Library Path</b>")),
+    WidgetCustomQt(PreferenceWidget::create_instance),
+
+    WidgetLabel(N_("<b>Playlist title</b>")),
+    WidgetEntry(N_("Library playlist"),
+		WidgetString(CFG_ID, "playlist_title")),
+    WidgetEntry(N_("Playing playlist"),
+		WidgetString(CFG_ID, "playing_title")),
+    WidgetSeparator(),
+	WidgetCheck(N_("Auto-send to playlist"), WidgetBool(CFG_ID, "auto_send"))
+};
+
+const PluginPreferences LibraryViewQt::prefs = {
+    {widgets}, PreferenceWidget::init, PreferenceWidget::apply, nullptr};
+
+LibraryTreeView* LibraryViewQt::get_tree_widget() { return tree_widget; }
+
+bool LibraryViewQt::init() {
+	aud_config_set_defaults(CFG_ID, defaults);
+	return true;
+}
+
+void* LibraryViewQt::get_qt_widget() {
+	if(!tree_widget)
+	tree_widget = new LibraryTreeView;
+
+	return tree_widget;
+}
+
+void set_pref_uri(QString filename) {
+    auto uri = filename_to_uri(
+	filename_expand(filename_normalize(str_copy(filename.toUtf8()))));
+    aud_set_str(CFG_ID, "library_path", uri);
+}
+
+String get_pref_uri() {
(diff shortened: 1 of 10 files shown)


jlindgren90:
Some thoughts/questions:

1. How many people have used this plugin so far and found it useful?
2. How is the performance with large and/or networked folders, or on a slow machine?
3. I think the name "Library View" (both for the plugin and the playlist it creates) could be confusing, especially since the "Library" analogue is already used by the Search Tool plugin. I would suggest naming the plugin "Folder Browser" and the playlist "Folder View".
4. I can't tell at a quick glance what the plugin is doing with the "Playing" playlist. But this also seems likely to be confusing since Audacious already uses a "Now Playing" playlist.
5. Please run clang-format on the source code to make the indentation more consistent.

@kaniini will probably make the final call on whether to merge this or not.

kaniini:
In general, I don't see the point over using the search-tool plugin for this.

To me, it seems like we could just add the drill-down feature to folder view, using the database built by search-tool.

kaniini:
To expand on what I mean: when using search tool, audacious already builds a very rich database using the Tuple system.

All the information is already there to replicate this functionality in search tool.  The drill-down concept is useful and would be welcome, but I don't think having a filesystem-specific view is actually helpful at all.  Most people just want to find and enqueue specific tracks, and search tool is already more useful for that unless you organize your music in a specific way.

So NAK from me for including this -- I don't see the point of it.  I think it would be great to add drill-down to search tool though.  That would really solve the problem that this plugin is trying to solve in a much more efficient way without forcing people to store their media in any specific directory layout.

src/library-view-qt/library-view-qt.cc
```diff
@@ -0,0 +1,83 @@
...
+    if(path[0] && test_uri(static_cast<const char*>(path)))
+	return path;
+    else
+	return to_uri(g_get_home_dir());
+}
+
+bool test_uri(QString path) {
+    StringBuf full_path = str_copy(path.toUtf8());
+    full_path = filename_expand(filename_normalize(uri_to_filename(full_path)));
+    return g_file_test(full_path, G_FILE_TEST_IS_DIR);
```
kaniini, line 70: please use `vfs_file_test`.  direct glib usage in audacious is mostly deprecated.

src/library-view-qt/library-view-qt.h
```diff
@@ -0,0 +1,57 @@
...
+#include <libaudcore/i18n.h>
+#include <libaudcore/playlist.h>
+#include <libaudcore/plugin.h>
+#include <libaudcore/preferences.h>
+
+#define CFG_ID "library-view"
+
+constexpr auto DEFAULT_LIBRARY_PATH = "~";
+constexpr auto DEFAULT_PLAYLIST_TITLE = "Library view";
+constexpr auto DEFAULT_PLAYING_TITLE = "Playing";
```
kaniini, line 17: If you use a temporary playlist, Audacious will already call it "Now playing".  Please do not add additional playlists for temporary use.

src/library-view-qt/library-view-qt.h
```diff
@@ -0,0 +1,57 @@
...
+
+#include <libaudcore/i18n.h>
+#include <libaudcore/playlist.h>
+#include <libaudcore/plugin.h>
+#include <libaudcore/preferences.h>
+
+#define CFG_ID "library-view"
+
+constexpr auto DEFAULT_LIBRARY_PATH = "~";
+constexpr auto DEFAULT_PLAYLIST_TITLE = "Library view";
```
kaniini, line 16: The search-tool plugin was expanded into being a full media library already, as noted in my general comments.  Please do not call this "library view".

src/library-view-qt/preference.cc
```diff
@@ -0,0 +1,79 @@
...
+	: QWidget(parent), box_layout(audqt::make_hbox(this)),
+	  file_entry(audqt::file_entry_new(this, _("Choose Folder"),
+					   QFileDialog::Directory,
+					   QFileDialog::AcceptOpen)) {
+	file_entry->setText(static_cast<const char*>(
+	filename_contract(uri_to_filename(get_pref_uri()))));
+	box_layout->addWidget(file_entry);
+	default_palette = file_entry->palette();
+	error_palette = default_palette;
+	error_palette.setColor(QPalette::Text, Qt::red);
```
kaniini, line 24: Errors are generally not colored in red like this in Audacious.  We prefer to use iconography, as it is more accessible.

mojyack:
Thank you for your comment.
Not many people may manage their music with directories, but for some people, the This feature is very useful.
We can actually find a several posts by people who need the feature to display files in their music player like a browser in a Google search.

I think it's a good idea to use the search-tool database, too.
I'm stuck on deciding whether to update the commented points without changing the features, or to change it to show the search-tool database in drill-down form.
If I choose the latter, I don't know if I should copy the source of search-tool into my plugin or if I should make it configurable as an option for search-tool.
Either way, I'm going to work a little
I'd be happy to give you some advice.

kaniini:
I think it would be better to enhance search-tool with drill-down.

mojyack:
OK, I'll leave this plugin aside for now and work on enhancing the search-tool. Thank you for your advice.

PR #92, opened 2020-06-30
Move from freedb.org to gnudb.gnudb.org

bket:
The freedb.org database of compact track listings has shut down. Any
program functionality that tries to fetch from or submit data to
freedb.org is broken now.

An alternative service is available at gnudb.gnudb.org, see
https://www.gnudb.org/

Fixes https://redmine.audacious-media-player.org/issues/994


jlindgren90:
gnudb.org looks like a good alternative. Thank you!

PR #94, opened 2020-09-04
Mergable gtk2 win hotkey

domenn:
Global Hotkeys support for gtk2 for Windows.

I have further ideas for improvements, hence some commented out experimental code. This is "checkpoint" state that works.

I am willing to keep supporting code of this plugin in future, in case other people have (and report) issues with it regarding my changes.


jlindgren90:
This would be really useful. I'm afraid I don't personally have time to give the code a good review right now.

domenn:
Thanks for initial positive response.
I am using audacious with those changes daily and it works without problems. I don't observe any regressions. Previously, there were issues with state not saving on close and things like that. But recently I put much higher effort into this to make it work without issues. 

I also tested on Linux to make sure it works there as before.

About code review, is there anyone else? Otherwise I hope that you can manage some moment in future. Hopefully before next release.

If you need some more information from me about why I did the things the way I did them, feel free to ask. And note that this is not the first PR from my side into Audacious.

jlindgren90:
Hi, I'm belatedly getting to look at this.
I ran clang-format on the master branch so that we can separate out the formatting changes from the function changes.
Can you please rebase on top of master like so:

```
git checkout master
git pull
git checkout mergableGtk2WinHotkey 
git rebase 7dbcee3faf027db690e3f3ea6afbeb7f2bb24187
clang-format -i src/hotkey/*.{cc,h}
git commit -a -m "Run clang-format on src/hotkey"
git branch rebaseGtk2WinHotkey 0a35159f8bfe8fe1d9a264fd3fb2ca21fdb18dbf
git checkout rebaseGtk2WinHotkey 
git checkout mergableGtk2WinHotkey -- .
clang-format -i src/hotkey/*.{cc,h}
git commit -a -m "Rebase Global Hotkeys support for gtk2 for Windows"
```

jlindgren90:
Also please remove commented-out code and the `#ifdef BUILT_FROM_CMAKE` sections.

domenn:
Done as you said.
I tried locally by my side, it compiles. I didn't test working, but it should work, because I tested last time when submitting PR. You test everything before you merge anyway, don't you?

jlindgren90:
Is there a way to handle global hotkeys without referencing the main Audacious window?

domenn:
As far as my knowledge goes, no. Windows API function [RegisterHotKey](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-registerhotkey) needs a window that is receiving messages. There may be a way to use some other window that is not "main Audacious window".

The way I did it needed significant research, and this is the method that I managed to come up with.

jlindgren90:
The documentation for RegisterHotKey says, "If this parameter [hWnd] is NULL, WM_HOTKEY messages are posted to the message queue of the calling thread and must be processed in the message loop." It looks like you would then retrieve those messages by passing hWnd = NULL again to GetMessage.

Please look into that method. The code in its current state is too tightly coupled to a specific UI plugin. Searching for the main window by title is particularly bad.

domenn:
I'm pretty busy these days. So, I cannot promise anything.

C-Duv:
> Is there a way to handle global hotkeys without referencing the main Audacious window?

Just to be sure, the related code is the following?:

```Cpp
bool WindowsWindow::is_main_window(bool allow_hidden) const
{
    if (main_window_hidden_ && !allow_hidden)
        return false;
    bool returning =
        (class_name_ == "gdkWindowToplevel" && [](const std::string & title) {
            return title == N_("Audacious") ||
                   title.find(translated_title()) != std::string::npos ||
                   title == N_("Buffering ..."); // msgid "%s - Audacious"
        }(win_header_));
    if (returning)
    {
        kind_ = AudaciousWindowKind::MAIN_WINDOW;
    }
    return returning;
}
```

([See in context](https://github.com/audacious-media-player/audacious-plugins/pull/94/files#diff-aca849cfe95ab60cdfd4ce2893a3fd2efeca31002064dd7bc2307e5c6304e63cR93-R108))

jlindgren90:
That's part of it. The whole `WindowsWindow` class should go away, I think.

Sokoloft:
Hello,

Sorry to necro. However I'd love this functionality too. Ideally using the media keys. With that, hopefully it'd be detected by KDE connect on my phone.

PR #95, opened 2020-12-22
Implement album artist support in scrobbler2 plugin

arcctgx:
The purpose of this PR is to add album artist support to `scrobbler2` plugin, with some additional fixes in functions used for handling scrobble cache file.

The first commit fixes two problems which can happen when the scrobble cache file (`scrobbler.log`) is parsed. In some cases reading incorrectly formatted fields in `scrobbler.log` can cause a segmentation fault (a minimal example is a field containing only three `\t` characters). The other problem is that reading an empty line from scrobble cache stops further cache processing, and causes all subsequent cache entries to be lost. 

This doesn't seem to be a problem in real life usage, because the plugin never creates incorrectly formatted cache entries. I only found these problems when I was feeding various incorrectly formatted cache files to the plugin while testing my changes. But IMO this is something that should be fixed anyway. And having a dedicated function for validating format made further work easier.

The second commit implements album artist support in the plugin. Album artist info is read from the metadata of media file, and is sent to Last.fm in `track.updateNowPlaying` and `track.scrobble` API method calls. Scrobble cache file format is changed to store an additional field, but the change is implemented so that it's backward-compatible (`scrobbler.log` created by previous plugin version  will still be parsed correctly).

### DIFF
--- a/src/scrobbler2/scrobbler.cc
+++ b/src/scrobbler2/scrobbler.cc
@@ -94,6 +94,7 @@ static gboolean queue_track_to_scrobble (void * data) {
     StringBuf artist = clean_string (playing_track.get_str (Tuple::Artist));
     StringBuf title  = clean_string (playing_track.get_str (Tuple::Title));
     StringBuf album  = clean_string (playing_track.get_str (Tuple::Album));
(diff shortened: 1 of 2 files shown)


jlindgren90:
Looks good to me. Thanks!

PR #97, opened 2021-02-03
xsf: update 2sf backend

ahigerd:
The 2sf implementation in the current plugin hasn't really been updated in a long time, but I picked up maintainership of the 2sf backend used by the Foobar2000 plugin a few months ago and I'm probably also going to be taking over the Winamp plugin.

Changes:
* Replaced the xSF parsing code with CyberBotX's implementation
* Updated DeSmuME to a more recent snapshot version
* Replaced DeSmuME SPU with a dedicated implementation for audio-only playback
* Fixed the default sample rate for more accurate playback
* Added support for ReplayGain and fade-out
* Added configuration parameters for sample rate, interpolation mode, and fade-out


src/xsf/XSFFile.cc
```diff
@@ -0,0 +1,285 @@
+/*
```
kaniini, line 1: There is no license on this file.

src/xsf/XSFFile.h
```diff
@@ -0,0 +1,61 @@
+/*
+ * xSF - File structure
+ * By Naram Qashat (CyberBotX) [cyberbotx@cyberbotx.com]
```
kaniini, line 3: Or this file.

kaniini (changes requested):

kaniini:
What are the advantages to replacing the xSF parser?  The one we already had was quite battle tested.  Without an actual explanation I reject that part of this MR.

src/xsf/XSFFile.cc
ahigerd: Oh, whoops. Thanks.

ahigerd:
The xSF loader you had couldn't handle libraries that themselves referenced other libraries. The parser also had a hard-coded list of supported tags and didn't expose a safe API for searching for other tags. It was wholly dependent on C-style manual memory management with hard-coded capacity limits despite this being a C++ project. The replacement implementation is also solidly battle-tested.

None of this was necessarily a deal-breaker and I acknowledge that I could have fixed them with sufficient effort, but since I'm also the maintainer of the 2sf CLI wave-rendering tool (and by extension the Foobar2000 plugin's backend) and I'm taking up the mantle of maintaining the 2sf Winamp plugin because CyberBotX doesn't want to do it anymore, it seemed to make more sense to centralize on an implementation that would be easier to keep in sync across those three projects.

src/xsf/XSFFile.cc
ahigerd: Turns out it's because CyberBotX never formally put a license on it. I talked to him and got his permission to apply BSD to reflect the influence from AO.

I also forgot to put license headers on the files I pulled from my own project. Fixed that too.

kaniini:
If you plan to take over the maintenance of the 2sf plugin, then it's fine to change the parser I suppose.

ahigerd:
I didn't want to make the assumption that I could just waltz in and take over the maintainership, but it is a responsibility I would be willing to pick up.

dakrk:
This has been something I've been waiting for for years, thank you so much
The current 2SF backend seems to have some sort of reverb issue and outright doesn't play certain ones (such as the Pokemon Black and White ones VixyNyan ripped), so this is perfect

radioactiveman:
Thanks for your contribution @ahigerd. :+1:

Comments:
- If GPU.cc/h are no longer required, remove them.
- Update also the meson.build file the next time. I did that via commit bc8b636020ce6220d8e2c36604261e8a95ccd6ee.
- There are a few compiler warnings now (see below).

GCC 10.2.0
```
../src/xsf/desmume/mc.cc: In member function 'bool BackupDevice::load_raw(const char*, uint32_t)':
../src/xsf/desmume/mc.cc:629:7: warning: ignoring return value of 'size_t fread(void*, size_t, size_t, FILE*)' declared with attribute 'warn_unused_result' [-Wunused-result]
  629 |  fread(&this->data[0], 1, size - left, inf);
      |  ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[56/389] Compiling C++ object src/xsf/xsf.so.p/desmume_MMU.cc.o
../src/xsf/desmume/MMU.cc: In function 'void MMU_Init()':
../src/xsf/desmume/MMU.cc:718:36: warning: 'void* memset(void*, int, size_t)' clearing an object of type 'struct MMU_struct' with no trivial copy-assignment; use assignment or value-initialization instead [-Wclass-memaccess]
  718 |  memset(&MMU, 0, sizeof(MMU_struct));
      |                                    ^
In file included from ../src/xsf/desmume/armcpu.h:23,
                 from ../src/xsf/desmume/NDSSystem.h:25,
                 from ../src/xsf/desmume/MMU.cc:25:
../src/xsf/desmume/MMU.h:301:8: note: 'struct MMU_struct' declared here
  301 | struct MMU_struct
      |        ^~~~~~~~~~
[58/389] Compiling C++ object src/xsf/xsf.so.p/desmume_SPU.cc.o
../src/xsf/desmume/SPU.cc: In function 'void SetDesmumeSampleRate(double)':
../src/xsf/desmume/SPU.cc:108:21: warning: unused variable 'chan' [-Wunused-variable]
  108 |     channel_struct *chan = &SPU_core->channels[i];
      |                     ^~~~
../src/xsf/desmume/SPU.cc: In function 'void SPU_SetSynchMode(int, int)':
../src/xsf/desmume/SPU.cc:218:5: warning: deleting object of abstract class type 'ISynchronizingAudioBuffer' which has non-virtual destructor will cause undefined behavior [-Wdelete-non-virtual-dtor]
  218 |     delete synchronizer;
      |     ^~~~~~~~~~~~~~~~~~~
```

Clang 11.0.1
```
../src/xsf/desmume/mc.cc:629:2: warning: ignoring return value of function declared with 'warn_unused_result' attribute [-Wunused-result]
        fread(&this->data[0], 1, size - left, inf);
        ^~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
[75/389] Compiling C++ object src/xsf/xsf.so.p/desmume_SPU.cc.o
../src/xsf/desmume/SPU.cc:108:21: warning: unused variable 'chan' [-Wunused-variable]
    channel_struct *chan = &SPU_core->channels[i];
                    ^
../src/xsf/desmume/SPU.cc:218:5: warning: delete called on 'ISynchronizingAudioBuffer' that is abstract but has non-virtual destructor [-Wdelete-abstract-non-virtual-dtor]
    delete synchronizer;
    ^
../src/xsf/desmume/SPU.cc:1047:10: warning: no case matching constant switch condition '-1'
  switch(CHANNELS)
         ^~~~~~~~
../src/xsf/desmume/SPU.cc:1073:7: note: in instantiation of function template specialization 'SPU_Mix<-1>' requested here
      SPU_Mix<CHANNELS>(SPU, chan, data);
      ^
../src/xsf/desmume/SPU.cc:1088:5: note: in instantiation of function template specialization '____SPU_ChanUpdate<0, -1>' requested here
    ____SPU_ChanUpdate<FORMAT,-1>(SPU,chan);
    ^
../src/xsf/desmume/SPU.cc:1101:13: note: in instantiation of function template specialization '___SPU_ChanUpdate<0>' requested here
    case 0: ___SPU_ChanUpdate<0>(actuallyMix, SPU, chan); break;
            ^
../src/xsf/desmume/SPU.cc:49:18: warning: unused function 'read_s8' [-Wunused-function]
static inline s8 read_s8(u32 addr) { return (s8)_MMU_read08<ARMCPU_ARM7,MMU_AT_DEBUG>(addr); }
                 ^
../src/xsf/desmume/SPU.cc:71:17: warning: unused variable 'indextbl' [-Wunused-const-variable]
static const s8 indextbl[8] = { -1, -1, -1, -1, 2, 4, 6, 8 };
                ^
../src/xsf/desmume/SPU.cc:73:18: warning: unused variable 'adpcmtbl' [-Wunused-const-variable]
static const u16 adpcmtbl[89] = {
                 ^
6 warnings generated.
```

ahigerd:
Oh, thanks for the heads up about Meson. I completely forgot about that, and thanks for taking care of it for me.

Comments are addressed in #98.

PR #100, opened 2021-02-09
skins-qt: fix menu inconsistencies

dakrk:
With the Winamp interface, Genre is before Release Date in the Sort Selected menu, whereas every other sort menu has Genre after Release Date


radioactiveman:
Merged, thanks for your contribution.

PR #101, opened 2021-02-10
skins-qt: support winamp theme titlebar dimming

dakrk:
Supports the titlebar dimming that Winamp themes have when the current window isn't focused.


jlindgren90:
Thanks! One suggestion: this dims the titlebar (of the main window) when you click on the equalizer or playlist window. It might be better to keep displaying the "active" titlebar when any of those three windows are focused, because keypress handling is shared between them.

kaniini:
I agree with @jlindgren90, please use active titlebar if the player is focused at all.

dakrk:
In regular Winamp, the main window gets dimmed whenever the equalizer or playlist is focused, which is the way it is implemented in this currently. I preferably want it to be more accurate to original Winamp, although if you prefer it the way you mentioned above I could have it be a setting, as your reasoning behind it does make sense.

kaniini:
A config setting would work for me.

dakrk:
Apologies for the almost 1 year delay

jlindgren90:
Nice work. Merged with 579936d7f2e806f410b0293b2f47f3a1f10ce043. Thanks!

PR #104, opened 2022-05-04
pipewire: Add new output plugin. Closes: #1048

radioactiveman:
This is my attempt to port the PipeWire output plugin from Qmmp to Audacious.
[Ilya Kotov](https://github.com/trialuser02), author of the related [Qmmp code](https://sourceforge.net/p/qmmp-dev/code/HEAD/tree/branches/qmmp-2.0/src/plugins/Output/pipewire/) which is under GPL, allowed me to use the BSD license for our code base.

I'm using the plugin successfully for some time now, but further tests are of course appreciated. :)
Could you please support with the remaining things listed below? I don't know how to do them.

* Implement `period_wait()`.
* Use the ring buffer from `libaudcore` instead of the raw char array.
* A few variables are still hard-coded, this should be cleaned-up.

Other questions:
- I assume the channel mappings in `set_channel_map()` can be reduced, since Audacious only supports up to [10 channels](https://github.com/audacious-media-player/audacious/blob/e7addd0068ac3cb17fbedc29bd852a53088184ad/src/libaudcore/audio.h.in#L23). Is this correct?
- Is it necessary to include all format mappings in `to_pipewire_format()`? The PulseAudio plugin for example only maps a few formats in `to_pulse_format()`. Why is that?
- Is the priority ok? Should it be lower than the PulseAudio plugin for now?


mcarans:
Would be great to see this merged.

PedroHLC:
Demo

LGTM, or better, sounding good to me (tested on NixOS).

jlindgren90:
The biggest issue I see here is `period_wait()` isn't implemented. Calling `pw_thread_loop_timed_wait()` from `write_audio()` will potentially block the UI thread and make Audacious as a whole less responsive. Until we find a way to solve that, I would think it would be better to continue using the PulseAudio output plugin, as I believe that should be usable still with PipeWire?

jlindgren90:
@radioactiveman It may enough to just move the `pw_thread_loop_timed_wait()` bit from `write_audio()` to `period_wait()`. Would you be able to make/test that change?  (I am not using PipeWire on any of my own systems currently.)

src/pipewire/pipewire.cc
```diff
@@ -0,0 +1,575 @@
...
+
+    // TODO: Don't hardcode values, use aud_get_int("output_buffer_size")?
+    //
+    // m_stride = AudioParameters::sampleSize(format) * map.count();
+    // m_frames = qBound(64, qCeil(2048.0 * rate / 48000.0), 8192);
+    // m_buffer_size = m_frames * m_stride;
+    // m_buffer = new unsigned char[m_buffer_size];
+    m_stride = 4;
+    m_frames = 128;
+    m_buffer_size = 1024;
```
jlindgren90, line 330: If the commented line above is correct
```
m_buffer_size = m_frames * m_stride;
```

then should this be:
```
m_buffer_size = 512;
```

?

src/pipewire/pipewire.cc
```diff
@@ -0,0 +1,575 @@
...
+    if (!(m_loop = pw_thread_loop_new("pipewire-main-loop", nullptr)))
+    {
+        AUDERR("PipeWireOutput: unable to create main loop\n");
+        return false;
+    }
+
+    if (!(m_context = pw_context_new(pw_thread_loop_get_loop(m_loop), nullptr, 0)))
+    {
+        AUDERR("PipeWireOutput: unable to create context\n");
+        return false;
```
jlindgren90, line 283: Most of the `return false;` paths in this function appear to leak memory/not clean things up properly (e.g. here we leak `m_loop` which was initialized above).

src/pipewire/pipewire.cc
```diff
@@ -0,0 +1,575 @@
...
+{
+    PipeWireOutput * o = static_cast<PipeWireOutput *>(data);
+
+    if (!o->m_buffer_at)
+    {
+        pw_thread_loop_signal(o->m_loop, false);
+        return;
+    }
+
+    struct pw_buffer * b = pw_stream_dequeue_buffer(o->m_stream);
```
jlindgren90, line 474: Can `pw_stream_dequeue_buffer()` return `nullptr`?  The line below will crash if so:
```
struct spa_buffer * buf = b->buffer;
```

jlindgren90 (changes requested):

radioactiveman:
Thanks for the review @jlindgren90 and thanks @PedroHLC for testing. :+1:
I will try to address the remarks in the next days when I have some time.

src/pipewire/pipewire.cc
radioactiveman: Fixed by calling `close_audio()` in these cases, similar to the PulseAudio plugin.

src/pipewire/pipewire.cc
```diff
@@ -0,0 +1,597 @@
...
+        values[0] = v.left / 100.0f;
+        values[1] = v.right / 100.0f;
+    }
+    else
+    {
+        for (unsigned int i = 0; i < m_channels; i++)
+            values[i] = aud::max(v.left, v.right) / 100.0f;
+    }
+
+    pw_thread_loop_lock(m_loop);
```
radioactiveman, line 153: Should the `lock` be done at the beginning of the method? Or does it not matter here? Same question for `init_stream()`.

src/pipewire/pipewire.cc
```diff
@@ -0,0 +1,597 @@
...
+    pw_thread_loop_signal(o->m_loop, false);
+}
+
+void PipeWireOutput::on_drained(void * data)
+{
+    PipeWireOutput * o = static_cast<PipeWireOutput *>(data);
+    pw_thread_loop_signal(o->m_loop, false);
+}
+
+enum spa_audio_format PipeWireOutput::to_pipewire_format(int format)
```
radioactiveman, line 535: Do we need to handle all formats here? The PulseAudio plugin checks less in [to_pulse_format()](https://github.com/audacious-media-player/audacious-plugins/blob/45a0df0ecb479a64d85f982b8cc087b7547e5562/src/pulse/pulse_audio.cc#L369).

radioactiveman:
> @radioactiveman It may enough to just move the `pw_thread_loop_timed_wait()` bit from `write_audio()` to `period_wait()`. Would you be able to make/test that change?

Yes, this works. :+1: And I have noticed that the volume slider in in fact more responsive now, especially with the GTK interface.

radioactiveman:
@jlindgren90: I have rebased this to `master` and addressed your review comments. CI is unfortunately not possible since the PipeWire package is too old on Ubuntu 20.04.

Could you please answer my two questions in the review? Thanks and looking forward to the merge. :)

src/pipewire/pipewire.cc
jlindgren90: I think they are fine where they are :+1:

src/pipewire/pipewire.cc
jlindgren90: I don't think we ever send 8-bit or unsigned formats to the output plugins. Supporting them doesn't do any harm though.

jlindgren90 (approved):

jlindgren90:
@radioactiveman Looks good to me. Thanks for all your work on this!

birdie-github:
PipeWire uses/offers _the exact same API_ as PulseAudio, in other words PA plugin is _100% compatible and sufficient_ for PipeWire.

There's _no_ need to have two plugins. You may rename the old one from "PulseAudio plugin" to "PulseAudio/PipeWire plugin" and that's it.

kaniini:
One plugin uses libpulse, and the other uses libpipewire directly.  There's no need to support either sound system, but we do anyway.

PR #105, opened 2022-05-09
Fix bitrate unit formatting

getsnoopy:
This commit updates the unit formatting for bitrates to the universal
symbol from the now deprecated abbreviation.


radioactiveman:
Thanks for your contribution. :+1:

radioactiveman:
@getsnoopy: There is one more label which uses "kb/s" instead of "kbit/s. I guess we should change it as well for consistency?

https://github.com/audacious-media-player/audacious/blob/33b0654e7f96be431d5ac3302b229e32361d7a87/src/libaudgui/infowin.cc#L469

PR #107, opened 2022-06-08
Fix sndfile plugin "Unknown sndfile"

bangun:
On system without ffmpeg, playing some audio format, for example Opus, will fallback to sndfile input plugin.
It work just fine, but it will show "Unknown sndfile" on status bar while playing these files.

This commit changes that to codec information, and should only affects previously unknown format and potentially new format supported by libsndfile in the future.


jlindgren90:
Looks good to me. Thanks!

PR #110, opened 2022-08-30
flac: Make Ogg FLAC support optional. Closes: #1181

radioactiveman:
If Ogg FLAC support is available depends on how the
FLAC library was compiled (-DWITH_OGG=ON/OFF).

See also: https://xiph.org/flac/api/group__flac__stream__decoder.html#details

### DIFF
--- a/src/flac/plugin.cc
+++ b/src/flac/plugin.cc
@@ -52,11 +52,20 @@ bool FLACng::init()
         eof_callback, write_callback, metadata_callback, error_callback,
         &s_cinfo);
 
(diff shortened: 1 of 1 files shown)


jlindgren90 (changes requested):
I think we should not bother creating the `ogg_flac_decoder` at all if `FLAC_API_SUPPORTS_OGG_FLAC` is `false`.

Other than that, LGTM.

jlindgren90:
I cherry-picked and cleaned up the decoder initialization with b40dcee6d595d293d8cfb6888f0b544bbfd283b3.  Thanks!

PR #111, opened 2022-09-07
Add publisher and catalog number (qtui, gtkui, flac, vorbis)

kramerc:
Requires audacious-media-player/audacious#53. Adds publisher and catalog number support to qtui, gtkui, flac, and vorbis.


jlindgren90:
Rebased and merged. Thanks!

PR #114, opened 2022-12-05
Add fractional scaling for skins

AlexJakeGreen:
Problem:
On HiDPI screens it is possible that UI is either microscopic (1x) or giant (2x). But an user is allowed to choose between 1x and 2x only.

With this change the user can set fractional scaling in order to fit the dimensions of his screen properly.

1x
1 6x


jlindgren90:
This has some issues:
- Scaling quality is very poor (looks like nearest neighbor)
- There are sometimes visible gaps between the playback buttons, depending on the scale factor
- It's difficult to type in the "Fractional scale" entry because the window loses focus after each digit typed

AlexJakeGreen:
* This is an upscale of 275pixel width image, and unfortunately, upscale quality is always poor by design. Good quality only possible if a skin is redrawn manually for a higher resolution, but all the skins available in the internet are of 275pixel width.
* Theoretically 1px fluctuations are possible due to conversion `double` → `int`. The gap obviously does not appear with coefficients `1.0`, `2.0` and `3.0`, which means the current behavior is not affected.
* I suspect this is how `WidgetFloat` is intended to work, it looks like people are expected to press the arrows ↑ and ↓ to set the float value.

jlindgren90:
No, no, and no:
- There are better upscaling algorithms available.
- The 1px gaps can be fixed - for example by more careful rounding, or by rendering the whole window at 1x before upscaling.
- WidgetFloat is intended to allow keyboard entry.

AlexJakeGreen:
I'm a bit confused. This PR is about allowing fractional scaling.
This PR does not change the algorithm of upscaling.
This PR does not change behavior of WidgetFloat.

Upscaling algorithm and WidgetFloat functionality is outside of scope of this PR.

AlexJakeGreen:
I'm not enough skilled to take additional work on fixing WidgetFloat and adding different upscaling algorithm, so I think it is better to close the PR since I can't proceed with additional work.

AlexJakeGreen:
Please delete this PR. I did not touch WidgetFloat but I experienced a push to fix it. This is not how the things are to be dealt. It's better if I stay away from here.

AlexJakeGreen:
@jlindgren90 
> There are better upscaling algorithms available.
Show me them in the code of the current repo

> The 1px gaps can be fixed
I can not reproduce. 

> WidgetFloat is intended to allow keyboard entry.
Are you pushing me to fix the code I did not touch?

AlexJakeGreen:
This PR needs to be deleted because it does not look like ready to be accepted here

PR #115, opened 2022-12-17
opus: Add new input plugin

radioactiveman:
So far Opus files were supported by ffaudio (via ffmpeg), now we support them also with a native Opus plugin.


radioactiveman:
Tested successfully with:
- Opus files converted from FLAC, including tags and album art
`opusenc --bitrate xxx input.flac output.opus`
- Opus streams from https://dir.xiph.org/codecs/Opus
and changing tags, e. g. http://stream.punkrockers-radio.de:8000/mobile.opus
- Opus files extracted from WebM files, downloaded with yt-dlp
`mkvextract tracks input.webm 0:output.opus`

Limitations:
- `write_tuple()` is not supported yet, but can follow later
- Only Ogg containers are supported, not WebM or MKV

See also:
- Documentation: https://opus-codec.org/docs/

Questions:
- Any code review comments?
- How large should be the buffer size? For now I use the same values as the Vorbis plugin. The [docs](https://opus-codec.org/docs/opusfile_api-0.12/group__stream__decoding.html#ga73493002d84a234f5e19b70f1dddbe2a) recommend this for `op_read_float(_buf_size)`. Is a PCM "frame" the same as a "sample"?
>  The number of floats that can be stored in _pcm. It is recommended that this be large enough for at least 120 ms of data at 48 kHz per channel (5760 samples per channel).

jlindgren90:
Thanks, I have been a bit busy lately but will try to take a look in the next week or two.

jlindgren90:
Very nice work, I have nothing to add :)

The buffer size probably doesn't matter very much (at least from our side, Audacious core / output plugins don't care).

PR #118, opened 2022-12-22
Restore GTK3 support

radioactiveman:
- Make use of the new libaudgui helpers
- Use recent GTK3 APIs if possible
- Refactor some methods slightly to avoid duplicate code


radioactiveman:
Pipeline is failing because this PR depends on https://github.com/audacious-media-player/audacious/pull/54.

jlindgren90:
I'm running short on time and only skimmed the diff, but didn't see anything that stood out. I'm okay with merging this (after https://github.com/audacious-media-player/audacious/pull/54).

PR #119, opened 2022-12-23
Select text already in SearchBar

Almoped:
If there is text in SearchBar, select it when the user presses ctrl+f.
If you use the mouse pointer you can still set the cursor where you want, without having to click twice.


jlindgren90:
Thanks. I found a simpler solution, see https://github.com/audacious-media-player/audacious-plugins/commit/33cf838f8d83b98b3a8eed576c8c64d144db0809.

PR #123, opened 2023-01-11
Fix Qt 6 deprecation warnings with a compatibility layer

radioactiveman:
For now it is sufficient to share this header internally.
In the future we may move it into libaudqt.


radioactiveman:
@jlindgren90: Is this fine for you or should this header be in libaudqt?

jlindgren90:
This is fine, we can always move the header later if we want to. Thanks!

PR #125, opened 2023-01-16
Update buildsys

radioactiveman:
Use the latest version from https://fl.nil.im/buildsys/timeline and make use of the new BUILDSYS_PLUGIN.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -19,10 +19,7 @@ AUD_COMMON_PROGS
 
 BUILDSYS_INIT
 BUILDSYS_SHARED_LIB
-
-if test $HAVE_DARWIN = yes ; then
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -19,10 +19,7 @@ AUD_COMMON_PROGS
 
 BUILDSYS_INIT
 BUILDSYS_SHARED_LIB
-
-if test $HAVE_DARWIN = yes ; then
-    BUILDSYS_FRAMEWORK([])
-fi
+BUILDSYS_PLUGIN
```
radioactiveman, line 23: @Midar: Is this the correct way and place for BUILDSYS_PLUGIN?

PR #126, opened 2023-01-17
Support new songlengths db format in SID plugin 

mvtiaine:
Ref https://www.hvsc.c64.org/download/C64Music/DOCUMENTS/Songlengths.faq

HVSC 68 introduced new songlengths db format.
Since HVSC 71 the old format db is no longer included.

`SidDatabase::length()` apparently assumes old db format while `SidDatabase::lengthMs()` works with new format, so switch to using that.
See: https://github.com/libsidplayfp/sidplayfp/blob/d241d27938b0f1f9f8a10d172eab1073868728fd/src/player.cpp#L597

Also:
* fix ROM file and the songlength db search path, which was broken due to missing / at least when using custom install prefix with Meson build
* require libsidplayfp >= 2.0 also with autotools build (Meson build already requires it), since the new db format is only supported with version >= 2.0.0

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -300,7 +300,7 @@ ENABLE_PLUGIN_WITH_DEP(sid,
     auto,
     INPUT,
     SIDPLAYFP,
-    libsidplayfp >= 1.0)
+    libsidplayfp >= 2.0)
 
 test_console () {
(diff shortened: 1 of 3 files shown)


jlindgren90:
Looks good. Thanks!

PR #127, opened 2023-02-09
amidiplug: Make use of audstring methods

radioactiveman:


radioactiveman:
Questions:

1. This fixes the following warning by cppcheck. Is it a false positive? The code in `commit()` works fine, but accessing index 0 in a string with length=0 seems wrong.
```
Out of bounds access in expression 'sflist_string[0]' because 'sflist_string' is empty.
```

2. The check for empty string in `update()` is no longer necessary in my opinion. Would you keep it anyway to avoid calling `str_list_to_index()`?

jlindgren90:
Looks good to me. I think you made the right call on both counts: the warning looks legitimate, and the check for empty string is no longer needed.

PR #131, opened 2023-03-08
Fix Qt 6 support in qthotkey

XRevan86:


src/qthotkey/meson.build
```diff
@@ -1,17 +1,18 @@
 if get_option('qt6')
-  qtx11extras_dep = dependency('qt6', version: qt_req, required: false, modules: ['X11Extras'])
+  qtprivate_dep = dependency('qt6', version: qt_req, private_headers: true, required: false, modules: ['Core', 'Gui'])
```
radioactiveman, line 2: We require Qt 6.0 globally but QX11Application exists only since 6.2. This version needs to be set explicitly here.

https://doc-snapshots.qt.io/qt6-dev/qnativeinterface-qx11application.html

src/qthotkey/plugin.cc
```diff
@@ -436,13 +440,27 @@ bool GlobalHotkeys::init()
...
     audqt::init();
 
-    if (!QX11Info::isPlatformX11())
+#ifndef Q_OS_UNIX
+    bool   native = false;
+#elif QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+    auto * native = qApp->nativeInterface<QNativeInterface::QX11Application>();
+#else
+    bool   native = QX11Info::isPlatformX11();
+#endif
```
radioactiveman, line 449: I would write it like that, without the indentation spaces and a properly named `bool` for all cases.

```c++
#ifndef Q_OS_UNIX
    bool has_x11 = false;
#elif QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
    QNativeInterface::QX11Application * x11_interface =
        qApp->nativeInterface<QNativeInterface::QX11Application>();
    bool has_x11 = x11_interface != nullptr;
#else
    bool has_x11 = QX11Info::isPlatformX11();
#endif
```

src/qthotkey/meson.build
```diff
@@ -1,17 +1,18 @@
...
 else
-  qtx11extras_dep = dependency('qt5', version: qt_req, required: false, modules: ['X11Extras'])
+  qtprivate_dep = dependency('qt5', version: qt_req, required: false, modules: ['X11Extras'])
 endif
 
-have_qthotkey = qtx11extras_dep.found()
+# Requires private headers and a UNIX system for X11 support
+have_qthotkey = qtprivate_dep.found() and target_machine.system() not in ['windows', 'darwin']
```
radioactiveman, line 8: You can reuse `have_darwin` and `have_windows` from the root meson.build file here.

radioactiveman (changes requested):
Please address my comments about a few minor issues. I see you have noticed and fixed the failing CI on macOS with Qt 6 already by yourself. :+1:

Thank you and Slava Ukraini. :ukraine:

src/qthotkey/plugin.cc
XRevan86: Yea, I was thinking about how badly I formatted that piece of code, took your version.

src/qthotkey/meson.build
```diff
@@ -1,17 +1,19 @@
...
-  qtx11extras_dep = dependency('qt6', version: qt_req, required: false, modules: ['X11Extras'])
+  # QX11Application requires QtGui private headers and Qt >= 6.2
+  qtx11_dep = dependency('qt6', version: '>= 6.2', private_headers: true, required: false, modules: ['Core', 'Gui'])
 else
-  qtx11extras_dep = dependency('qt5', version: qt_req, required: false, modules: ['X11Extras'])
+  qtx11_dep = dependency('qt5', version: qt_req, required: false, modules: ['X11Extras'])
 endif
 
-have_qthotkey = qtx11extras_dep.found()
+have_qthotkey = qt_dep.found() and x11_dep.found() and qtx11_dep.found()
```
radioactiveman, line 9: We require `qt_dep` in the root meson.build file already, you can omit the check here.

src/qthotkey/meson.build
```diff
@@ -1,17 +1,19 @@
+x11_dep = dependency('x11-xcb', required: false)
```
radioactiveman, line 1: Nice approach, it's better than relying on `host_machine.system()`. :+1:
I would check for `x11` though. Any reason to use `x11-xcb` here?

radioactiveman (changes requested):

src/qthotkey/meson.build
XRevan86: Firstly, qthotkey [has an import from xcb](https://github.com/audacious-media-player/audacious-plugins/blob/audacious-plugins-4.3/src/qthotkey/plugin.cc#L59).  
Secondly, Qt [implements](https://doc.qt.io/qt-6/linux-requirements.html) X11 compatibility through xcb.

XRevan86: But x11-xcb was still found on macOS, so it seems `host_machine.system()` is inevitable.

radioactiveman (approved):

radioactiveman:
Merged now, thank you again @XRevan86.

PR #132, opened 2023-03-11
meson: Improve library detection

madpilot78:
Fix detection of some libraries in FreeBSD

- Use `pkgconfig` to detect `faad` (using `faad2.pc`)
- Look for `mp3lame` and `lirc_client` in `/usr/local/lib` on FreeBSD
- Remove unneeded `link_args: ['-llirc_client']`, required arguments provided by `cxx.find_library()`

### DIFF
--- a/src/aac/meson.build
+++ b/src/aac/meson.build
@@ -1,5 +1,10 @@
-faad_dep = cxx.find_library('faad', required: false)
-have_aac = faad_dep.found() and cxx.has_header('neaacdec.h')
+faad_dep = dependency('faad2', required: false)
+
+if not faad_dep.found()
+  faad_dep = cxx.find_library('faad', has_headers: 'neaacdec.h',
+                              dirs: ['/usr/local/lib'], required: false)
+endif
+have_aac = faad_dep.found()
 
 
 if have_aac
--- a/src/aac/meson.build
+++ b/src/aac/meson.build
@@ -2,7 +2,7 @@ faad_dep = dependency('faad2', required: false)
 
 if not faad_dep.found()
   faad_dep = cxx.find_library('faad', has_headers: 'neaacdec.h',
-                              dirs: ['/usr/local/lib'], required: false)
+                              required: false)
 endif
 have_aac = faad_dep.found()
 
--- a/src/filewriter/meson.build
+++ b/src/filewriter/meson.build
@@ -34,9 +34,14 @@ endif
 
 
 if get_option('filewriter-mp3')
-  lame_dep = cxx.find_library('mp3lame', required: false)
+  lame_dep = dependency('mp3lame', required: false)
 
-  if lame_dep.found() and cxx.has_header('lame/lame.h')
+  if not lame_dep.found()
+    lame_dep = cxx.find_library('mp3lame', has_headers: 'lame/lame.h',
+                                dirs: ['/usr/local/lib'], required: false)
+  endif
+
+  if lame_dep.found()
     filewriter_deps += [lame_dep]
     filewriter_srcs += ['mp3.cc']
 
--- a/src/filewriter/meson.build
+++ b/src/filewriter/meson.build
@@ -34,7 +34,7 @@ endif
 
 
 if get_option('filewriter-mp3')
-  lame_dep = dependency('mp3lame', required: false)
+  lame_dep = dependency('lame', required: false)
 
   if not lame_dep.found()
     lame_dep = cxx.find_library('mp3lame', has_headers: 'lame/lame.h',
--- a/src/filewriter/meson.build
+++ b/src/filewriter/meson.build
@@ -38,7 +38,7 @@ if get_option('filewriter-mp3')
 
   if not lame_dep.found()
     lame_dep = cxx.find_library('mp3lame', has_headers: 'lame/lame.h',
-                                dirs: ['/usr/local/lib'], required: false)
+                                required: false)
   endif
 
   if lame_dep.found()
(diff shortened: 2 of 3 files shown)


radioactiveman:
1. This has nothing to do with libintl, you only copy-pasted the description from your other pull request.
2. Is the correct name `lirc` or `lirc_client`?
3. Could we use `pkgconfig` (using `dependency()` instead of `cxx.find_library()` on FreeBSD? I don't like workarounds like these and would prefer a general solution. If needed with a fallback to the current approach to support older distributions.

madpilot78:
@radioactiveman 

> 1. This has nothing to do with libintl, you only copy-pasted the description from your other pull request.
> 

Ops sorry, I prepared the commit messages in a separate editor, and must have pasted the wrong text there. I'll fix it later while rebasing this patch.

>     2. Is the correct name `lirc` or `lirc_client`?

In FreeBSD the library is installed as `/usr/local/lib/liblirc_client.so` by default. so I guess correct one is `lirc_client`. I think this is due to the library providing only client utilities.

>     3. Could we use `pkgconfig` (using `dependency()` instead of `cxx.find_library()` on FreeBSD? I don't like workarounds like these and would prefer a general solution. If needed with a fallback to the current approach to support older distributions.

In this case I used `cxx.find_library()` since that was what was being used, I can change my patch easily.

As far as I can see `lirc_client` does not provide a `.pc` file, so it cannot be detected via pkgconfig.

OTOH faad does provide a `.pc` file, so I can use `dependency()` for that.

>     4. The `filewriter` plugin also uses `cxx.find_library()` to search `lame`. Does this work on FreeBSD?

I admit I did not check this, and it is actually failing to find it, so I will also fix this one.

This library is also not providing a `.pc` file, so I will need to keep using `cxx.find_library()`.

I'll rebase this patch adding the required fixes.

Thanks for you feedback!

madpilot78:
Regarding faad, I guess the patch can be generalized and the same detection used for all OSes. I'm unable to test this properly though.

radioactiveman:
I have checked the packages on Arch Linux. All of them ([faad2](https://archlinux.org/packages/extra/x86_64/faad2/files/), [LIRC](https://archlinux.org/packages/extra/x86_64/lirc/files/), [lame](https://archlinux.org/packages/extra/x86_64/lame/files/)) ship a pkgconfig file.

My idea was to use an approach like shown below. First search with `pkgconfig`, then with `find_library()` as fallback. The `dirs` parameter can be passed generally. The default dir (like /usr/lib) is then still searched. And the header check can also be included here since we require Meson 0.51 meanwhile.

See also: https://mesonbuild.com/Reference-manual_returned_compiler.html#compilerfind_library

Would you do this for all three libraries? Also include a "meson: " prefix in the commit message. Thanks.
Please test on FreeBSD, our CI does the rest. :smiley:

```patch
diff --git a/src/lirc/meson.build b/src/lirc/meson.build
index 9d214bb31..475f8b00d 100644
--- a/src/lirc/meson.build
+++ b/src/lirc/meson.build
@@ -1,13 +1,17 @@
-lirc_dep = cxx.find_library('lirc', required: false)
-have_lirc = lirc_dep.found() and cxx.has_header('lirc/lirc_client.h')
+lirc_dep = dependency('lirc', required: false)
 
+if not lirc_dep.found()
+  lirc_dep = cxx.find_library('lirc_client', has_headers: 'lirc/lirc_client.h',
+                              dirs: ['/usr/local/lib'], required: false)
+endif
+
+have_lirc = lirc_dep.found()
 
 if have_lirc
   shared_module('lirc',
     'lirc.cc',
     dependencies: [audacious_dep, glib_dep, lirc_dep],
     name_prefix: '',
-    link_args: ['-llirc_client'],
     install: true,
     install_dir: general_plugin_dir
   )
```

madpilot78:
@radioactiveman Good idea!

As far as I can see the lirc project distribution itself does not include a pc file, and the FreeBSD port is not adding one.

Anyway, sure I'll update the patch with the suggested behavior.

radioactiveman:
1. The pkgconfig name is `lame`, not `mp3lame`.
2. Nit-pick but I would add an empty line between the `endif` and `have_[aac,lirc]`. :wink:
3. Meson supports `LIBRARY_PATH`. Does this work for you on FreeBSD?
Removing the `dirs: ['/usr/local/lib']` parameters again would fix the failing Windows CI build.


```sh
export LIBRARY_PATH=/usr/local/lib
meson setup build
meson compile -C build
```

jlindgren90:
I would much prefer the `LIBRARY_PATH` approach of that works. Hard-coding `/usr/local` in a bunch of places seems wrong.

madpilot78:
> 1. The pkgconfig name is `lame`, not `mp3lame`.

I was not aware of this. I'll change that.

> 
>     2. Nit-pick but I would add an empty line between the `endif` and `have_[aac,lirc]`. wink

:white_check_mark:

> 
>     3. Meson supports `LIBRARY_PATH`. Does this work for you on FreeBSD?
>        Removing the `dirs: ['/usr/local/lib']` parameters again would fix the failing Windows CI build.
> 
> 
> ```shell
> export LIBRARY_PATH=/usr/local/lib
> meson setup build
> meson compile -C build
> ```

Should work in FreeBSD, I need to do some testing though, but it does look cleaner on my side too, since, in the FreeBSD ports tree, /usr/local/can be changed by the user a t build time and I have been using substitutions for now.

I'll also update audacious/pull#60 ro use `LIBRARY_PATH` if it works.

madpilot78:
@radioactiveman @jlindgren90 

I confirm using LIBRARY_PATH works fine for me, so I have updated the patch.

jlindgren90:
Assuming CI passes, this looks good to me.

radioactiveman:
Pushed as https://github.com/audacious-media-player/audacious-plugins/commit/bf40a4b8fa7d8e0670a4071edf8950f75d37f48c with minor clean-ups. Thanks again @madpilot78 for your contribution.

PR #133, opened 2023-03-21
mpg123: avoid false positives with content based detection

mvtiaine:
mpg123 detection has quite high false positive rate  when doing content based probing. This PR fixes at least some of them.

The actual improvement (in my testing) is due to this commit: https://github.com/audacious-media-player/audacious-plugins/pull/133/commits/a8933a05052ea2431ede48a86c5217c6adfa2eb4

Fixes false positives for example among these files: 
(MP2 falsely detected): http://ftp.modland.com/pub/modules/Delitracker%20Custom/TSM/donner/instr/sphere.x
(MP3 falsely detected): http://ftp.modland.com/pub/modules/Delitracker%20Custom/TSM/donner/instr/donner.x
More examples: https://ftp.modland.com/pub/modules/Zoundmonitor/Samples/

Also fixes https://redmine.audacious-media-player.org/issues/1188

### DIFF
--- a/src/mpg123/mpg123.cc
+++ b/src/mpg123/mpg123.cc
@@ -168,7 +168,7 @@ DecodeState::DecodeState(const char * filename, VFSFile & file, bool probing,
     if (mpg123_open_handle(dec, &file) < 0)
         goto err;
 
-    if (!stream && aud_get_bool("mpg123", "full_scan") && mpg123_scan(dec) < 0)
+    if (!stream && !probing && aud_get_bool("mpg123", "full_scan") && mpg123_scan(dec) < 0)
         goto err;
 
     while (1)
--- a/src/mpg123/mpg123.cc
+++ b/src/mpg123/mpg123.cc
@@ -54,7 +54,10 @@ public:
     static constexpr PluginInfo info = {N_("MPG123 Plugin"), PACKAGE, nullptr,
                                         &prefs};
     static constexpr InputInfo iinfo =
-        InputInfo(FlagWritesTag).with_exts(exts).with_mimes(mimes);
+        InputInfo(FlagWritesTag)
+            .with_exts(exts)
+            .with_mimes(mimes)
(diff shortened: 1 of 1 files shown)


jlindgren90:
mpg123 supports MPEG-1 layer 1, 2, or 3 (hence the "123") so rejecting layer 2 files is not right.

mvtiaine:
I did test with the few MP2 files I had that they still get detected even when removing the .mp2 extension.

mvtiaine:
I removed the layer <= 2 check from heuristic, as just noticed there were also MP3 false detections for example for this http://ftp.modland.com/pub/modules/Delitracker%20Custom/TSM/donner/instr/donner.x

mvtiaine:
Fixes also https://redmine.audacious-media-player.org/issues/1188

jlindgren90:
Makes sense, I misunderstood at first what the "layer 2" part was doing.

> The actual improvement (in my testing) is due to this commit: https://github.com/audacious-media-player/audacious-plugins/commit/a8933a05052ea2431ede48a86c5217c6adfa2eb4

If this is the only commit that shows actual improvement, then can we just merge it on its own? Or does it depend on the others?

mvtiaine:
Yes merging just that is fine with me. I double checked all the linked issues are also fixed by that alone.
The others are things I tried first and included as they seemed either real fixes or relevant optimizations.

Commits are independent, except for https://github.com/audacious-media-player/audacious-plugins/pull/133/commits/326015b43216a62e643ee7322c0043a865f8cdd6 which depends on https://github.com/audacious-media-player/audacious-plugins/pull/133/commits/1724f2af3ce732e74d91084316fe2931380298f9

jlindgren90:
I cherry-picked the one commit with https://github.com/audacious-media-player/audacious-plugins/commit/0abda62588a8ddaa49c98d412186ceacdefcc7b6. Thanks!

PR #134, opened 2023-03-31
Add a WM_WINDOW_ROLE property to toplevel X11 windows.

gijsbers:
With window managers like icewm and i3 users can configure window specific settings. For this to work toplevel windows must be distinguishable by a X11 property. Because Audacious uses the same WM_CLASS property for all its windows, the WM_WINDOW_ROLE property must be set. However, Audacious currently does not set this property. It is defined in the ICCCM manual for X11 clients. This patch sets this property on Gtk and Qt toplevel windows using designated library calls.

### DIFF
--- a/src/amidiplug/i_fileinfo.cc
+++ b/src/amidiplug/i_fileinfo.cc
@@ -152,6 +152,7 @@ bool i_fileinfo_gui (const char * filename_uri, VFSFile & file)
     /*****************************************************/
 
     fileinfowin = gtk_window_new (GTK_WINDOW_TOPLEVEL);
+    gtk_window_set_role (GTK_WINDOW (fileinfowin), "fileinfo");
     gtk_window_set_default_size (GTK_WINDOW (fileinfowin), 500, 400);
     gtk_window_set_type_hint (GTK_WINDOW (fileinfowin), GDK_WINDOW_TYPE_HINT_DIALOG);
     g_signal_connect (G_OBJECT (fileinfowin), "destroy", G_CALLBACK (gtk_widget_destroyed), &fileinfowin);
@@ -348,6 +349,7 @@ bool i_fileinfo_gui (const char * filename_uri, VFSFile & file)
 
     title = g_path_get_basename (filename_utf8);
     gtk_window_set_title (GTK_WINDOW (fileinfowin), title);
+    gtk_window_set_role (GTK_WINDOW (fileinfowin), "fileinfo");
     g_free (title);
     /* set the text for the filename header too */
     gtk_entry_set_text (GTK_ENTRY (title_name_v_entry), filename_utf8);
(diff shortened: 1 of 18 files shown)


src/skins-qt/playlistwin.cc
```diff
@@ -476,6 +476,7 @@ static void playlistwin_create_window ()
 
     playlistwin = new PlWindow (shaded);
     playlistwin->setWindowTitle (_("Audacious Playlist Editor"));
+    playlistwin->setWindowRole ("editor");
```
jlindgren90, line 479: `playlist` would be better than `editor` IMO.

src/skins/playlistwin.cc
```diff
@@ -489,6 +489,7 @@ static void playlistwin_create_window ()
 
     playlistwin = new PlWindow (shaded);
     playlistwin->setWindowTitle (_("Audacious Playlist Editor"));
+    playlistwin->setWindowRole ("editor");
```
jlindgren90, line 492: `playlist` here too.

src/skins/window.cc
```diff
@@ -93,6 +93,7 @@ Window::Window (int id, int * x, int * y, int w, int h, bool shaded) :
     h *= config.scale;
 
     GtkWidget * window = gtk_window_new (GTK_WINDOW_TOPLEVEL);
+    gtk_window_set_role ((GtkWindow *) window, "skin");
```
jlindgren90, line 96: The role should be set by the subclasses (`mainwindow` or `playlist` or `equalizer`) rather than by the base class.

jlindgren90 (commented):
Some of the window roles don't seem correct; see individual comments.

Also it looks like both `skins` and `skins-qt` are missing calls to set the window role for the main window.

src/gtkui/ui_gtk.cc
```diff
@@ -954,6 +954,7 @@ bool GtkUI::init ()
     menu_tab = make_menu_tab (accel);
 
     add_dock_plugins ();
+    gtk_window_set_role ((GtkWindow *) window, "mainwindow");
```
radioactiveman, line 957: This looks out of place here. Why not move it to line 801 where `window` is initialized?

src/amidiplug/i_fileinfo.cc
```diff
@@ -348,6 +349,7 @@ bool i_fileinfo_gui (const char * filename_uri, VFSFile & file)
 
     title = g_path_get_basename (filename_utf8);
     gtk_window_set_title (GTK_WINDOW (fileinfowin), title);
+    gtk_window_set_role (GTK_WINDOW (fileinfowin), "fileinfo");
```
jlindgren90, line 352: It looks like `gtk_window_set_role()` is called twice for the same window (here and line 155).

src/skins/window.cc
jlindgren90: This still needs to be fixed.

jlindgren90 (changes requested):
Still looks like both `skins` and `skins-qt` are missing calls to set the window role for the main window.

jlindgren90:
Thanks!

PR #136, opened 2023-04-22
flac: Check for seek errors. Closes: #1211

radioactiveman:


radioactiveman:
@jlindgren90: This was reproducible with the sample file linked in [#1211](https://redmine.audacious-media-player.org/issues/1211). Seeking fails when trying to seek exactly to the end. This obviously causes an error message now. Using parentheses would avoid this error message, but is less precise.

What's your recommendation?

```c++
(uint64_t) seek_value * s_cinfo.sample_rate / 1000
```
->
```c++
(uint64_t) seek_value * (s_cinfo.sample_rate / 1000)
```

jlindgren90:
> Using parentheses would avoid this error message, but is less precise.

Right, we don't want to add parentheses here, that would basically truncate any fractional part of the seek time and allow seeking only to whole-second times.

I expect the root cause of the error is trying to seek to a sample >= the length of the file? If that's the case, and if we're able to reliably know the length of the file on samples, maybe we could clamp the computed sample value to avoid that.

Detecting the error is a good start though. Happy to merge this as is.

radioactiveman:
I have debugged it further, your assumption is correct. :+1:
The error comes from [this check](https://github.com/xiph/flac/blob/1.4.2/src/libFLAC/stream_decoder.c#L1132-L1133) where `sample == total_samples` is true for the example file.

How would you clamp the sample value?

radioactiveman:
@jlindgren90: I came up with this patch, including a typo fix in the debug message. Please review.

```diff
diff --git a/src/flac/plugin.cc b/src/flac/plugin.cc
index ca013d098..d59d0069d 100644
--- a/src/flac/plugin.cc
+++ b/src/flac/plugin.cc
@@ -167,8 +167,13 @@ bool FLACng::play(const char *filename, VFSFile &file)
         int seek_value = check_seek ();
         if (seek_value >= 0)
         {
-            if (! FLAC__stream_decoder_seek_absolute (decoder, (uint64_t)
-             seek_value * s_cinfo.sample_rate / 1000))
+            uint64_t sample = (uint64_t) seek_value * s_cinfo.sample_rate / 1000;
+
+            /* Avoid error when seeking to a sample >= total_samples */
+            if (s_cinfo.total_samples > 0)
+                sample = aud::min(sample, s_cinfo.total_samples - 1);
+
+            if (! FLAC__stream_decoder_seek_absolute(decoder, sample))
             {
                 AUDERR("Error while seeking!\n");
                 error = true;
diff --git a/src/flac/seekable_stream_callbacks.cc b/src/flac/seekable_stream_callbacks.cc
index 8b3e7aaa5..d733d7bda 100644
--- a/src/flac/seekable_stream_callbacks.cc
+++ b/src/flac/seekable_stream_callbacks.cc
@@ -150,7 +150,7 @@ void metadata_callback(const FLAC__StreamDecoder *decoder, const FLAC__StreamMet
     if (metadata->type == FLAC__METADATA_TYPE_STREAMINFO)
     {
         info->total_samples = metadata->data.stream_info.total_samples;
-        AUDDBG("total_total_samples=%ld\n", (long) metadata->data.stream_info.total_samples);
+        AUDDBG("total_samples=%ld\n", (long) metadata->data.stream_info.total_samples);
 
         info->bits_per_sample = metadata->data.stream_info.bits_per_sample;
         AUDDBG("bits_per_sample=%d\n", metadata->data.stream_info.bits_per_sample);

```

jlindgren90:
> @jlindgren90: I came up with this patch, including a typo fix in the debug message. Please review.

Looks great, please go ahead and push 👍

PR #137, opened 2023-04-23
psf: Use snprintf instead of insecure sprintf

radioactiveman:
sprintf/vsprintf are deprecated since macOS 13.


jlindgren90:
Looks good, thanks!

PR #138, opened 2023-05-02
Add Playback History Qt-only plugin

vedgy:
Copy and modify the playlist-manager-qt plugin code.

A similar feature was requested in #553. However, this initial version of the playback history plugin tracks and displays only album history, not individual song/track history. Playback start/stop time is not tracked or shown either.


jlindgren90:
Looks like a nice plugin, except some of the limitations seem quite significant:

> tracks and displays only album history, not individual song/track history

> History entries are stored only in memory and are lost on Audacious exit.

I wonder how many users actually want a plugin that remembers only album history, and forgets it on exit?

vedgy:
My use case is listening to music while not otherwise using computer (the display is off). Then being able to tell what was playing. Storing and loading history is useless then.


Implementing storing and loading in an ini file is relatively easy. So if someone needs it, they can jump in. But there is a complication of automatic removal of old entries to limit the growth of history.

Tracking individual songs calls for automatic history cleanup as well, because the history will grow faster. Plus plugin preferences has to be implemented to choose between album and track history.

I don't want to implement either feature, because I'm not going to use them and can't be sure what exact behavior would be desired by the users who need these features.

vedgy:
> I wonder how many users actually want a plugin that remembers only album history, and forgets it on exit?

While it's not exactly what many people want, the feature can be somewhat useful to random-song users too. The playlist Entry Number of a first played song in an album is stored. Selecting a history entry selects this song in the playlist. Assuming the playlist is unmodified, one can follow the played songs. Except when consecutive songs happen to be picked from a common album.

jlindgren90:
Sorry for the delayed response, but I just haven't had time to review this so far.

vedgy:
No problem. I'll wait for a review.

vedgy:
Resolved trivial conflicts and rebased on *master*. Will integrate relevant changes since my last testing, if any, from the copied-and-modified plugin *playlist-manager-qt*. Then perhaps add the discussed song history option.

vedgy:
Made a few changes that do not require retesting to the first 3 commits. Implemented the Song mode and preferences to switch between the Song and the Album modes. Tested everything again. Please review.

I hope permanent history storage is not a requirement for merging. Implementing permanent storage would require adding a history size limit option and implementing automatic history cleanup. Maybe also changing the type of `m_entries` from `Index<HistoryEntry>` to `QList<HistoryEntry>` and declaring `HistoryEntry` as `Q_RELOCATABLE_TYPE` in order to support efficient removal from the front.

By the way, perhaps it makes sense to declare `HistoryEntry` as `Q_RELOCATABLE_TYPE` in the first existing commit just in case it's needed in the future?

radioactiveman:
> @radioactiveman can you review this? Not sure when I will be able to get to it.

Yes, I will try out the plugin and review its code when finding some time.

radioactiveman:
Thanks @vedgy for your nice work on the new plugin.
I pushed some additional minor changes and merged it all now. Sorry that it took so long.

Happy New Year to you and Slava Ukraini. :ukraine:

src/skins-qt/menus.cc
```diff
@@ -90,7 +90,7 @@ static const audqt::MenuItem main_items[] = {
 
 static const audqt::MenuItem playback_items[] = {
     audqt::MenuCommand ({N_("Song Info ..."), "dialog-information", "I"}, audqt::infowin_show_current),
-    audqt::MenuCommand ({N_("Playback History ..."), "view-history", "H"}, action_playback_history),
+    audqt::MenuCommand ({N_("Playback History ..."), nullptr, "H"}, action_playback_history),
```
vedgy, line 93: `locate view-history` finds the `view-history` icon in 3 icon themes on my system: _breeze_, _breeze-dark_ and _oxygen_. What happens when the icon is missing, as I assume is the case on your system?

radioactiveman: Then this icon seems to be KDE specific. When it is missing no icon is shown.

radioactiveman: Fair enough, please check my two last commits from my [fork](https://github.com/radioactiveman/audacious-plugins/commits/master/). Can I merge them into the main repo?

vedgy: Both commits look good, thank you. Please merge.

By the way, defining [`QT_NO_CAST_FROM_ASCII`](https://doc-snapshots.qt.io/qt6-dev/qstring.html#QT_NO_CAST_FROM_ASCII) project-wide is considered a good Qt practice. KDE projects define it [by default](https://api.kde.org/ecm/kde-module/KDECompilerSettings.html#modern-code).

PR #140, opened 2023-06-05
Adding "radio station" like transitions.

giantg:
Hi,

First let me say a huge thank you to the team that developed/maintains audacious!  This really seems to be a simple no-nonsense light weight audio player that "just works" (and on many platforms!)

I would call this pull request somewhat of a "work in progress" but I wanted to create a pull request to gauge interest as well as get feedback.

After tinkering with software another passion of mine is music and DJing.  I've been searching for years for an audio player that could do what the old "foo_continuator" plugin did for Foobar2000.  For years I've used an ancient version of Foobar2000 and that plugin to play "background music" or "cocktail hour" music at DJ gigs during the portions of the event that were not "dancing/high energy." 

It seems just about all audio players out there seem to implement "crossfade" the same way as a true X-fade.  While crossfading does happen when a DJ is (so-called "beatmaching") and mixing two songs keeping the rhythms together it usually does *NOT* when one is just playing back-to-back songs the way a broadcast radio station does.   From my experience, broadcast stations typically start the next program at full volume while fading out the previous program (radio automation software even usually stores an EOP - End of Program or EOM - End of Message tag in the media files telling the automation software exactly where to start the next program and start fading the previous).

My changes are probably a bit "hacky" as I'm brand new to the code base (please point out any places I've done something silly or dumb).  I believe the original functionality is preserved, but I've added a few new parameters to the configuration:

-  No Fade In:  The main reason for this change.  This boolean will instruct the plugin to NOT use the ramp/fade-in function for the incoming (next) program.  However, the ramp/fade-out function will still be used for the outgoing (current) program.

- Use S-Curve for fades:  Upon my initial listening tests against various types of musical programs (mostly commercial releases from the last 6 or so decades) the linear fade-out sounded "a little off" to my ears.  If I got my math right this should apply a Sigmoid or (S-Curve) function to the fade function: 1/(1+EXP(&lt;steepness&gt;\*(2\*i-length)/length))
-  S-Curve Steepness: Changes the &lt;steepness&gt; parameter in the sigmoid function above.  A lower number means a slower fade and higher is faster. This parameter does not affect the original linear style fades.
- Override Fade Out Length:  Included to allow for fade-outs shorter than the total "Overlap" length.  I'm not sure if this needs to stay as in my later testing the "Overlap" parameter was enough "configurability" to control the character of the transitions.
- Fade Out Length (seconds):  The length of the fade-out if "Override Fade Out Length" is set.

Am I going about this the correct/desired way or would this be better suited as a separate plugin?  I figured the crossfade plugin was somewhat "in charge" of program transitions so I figured the changes would be best made here.

One last thought:  Another consideration I had while working on these changes was implementing the reading of "Mixramp" tags (if they existed) in each media file and allow those tags to give fine or exact control over where the transition would occur (similar to the PROGRAM START/PROGRAM END tags used in radio automation software).  This would allow one to control the transition on specific files.  One use case might be a program or song that ends with a few quiet lyrics.  I've noticed that the plugin's transition can kind of "step on" those last few quiet lyrics at the end of a song that has them.

Sorry for the super long pull request message.  Feel free to tell me I'm going overboard on the "transitions" aspect of Audacious! :-)

Thanks again for a great audio player!

### DIFF
--- a/src/crossfade/crossfade.cc
+++ b/src/crossfade/crossfade.cc
@@ -21,6 +21,7 @@
 #include <libaudcore/plugin.h>
 #include <libaudcore/preferences.h>
 #include <libaudcore/runtime.h>
+#include <bits/stdc++.h>
 
 enum
 {
@@ -33,9 +34,14 @@ enum
 
 static const char * const crossfade_defaults[] = {
     "automatic", "TRUE",
-    "length", "5",
+    "length", "6",
     "manual", "TRUE",
     "manual_length", "0.2",
+    "no_fade_in", "TRUE",
+    "fade_out_override", "FALSE",
+    "fade_out_override_length", "4",
+    "use_sigmoid", "TRUE",
+    "sigmoid_steepness", "6",
     nullptr
 };
 
@@ -56,6 +62,20 @@ static const PreferencesWidget crossfade_widgets[] = {
     WidgetSpin (N_("Overlap:"),
         WidgetFloat ("crossfade", "manual_length"),
         {0.1, 3.0, 0.1, N_("seconds")},
+        WIDGET_CHILD),
+    WidgetCheck (N_("No fade in"),
+        WidgetBool ("crossfade", "no_fade_in")),
+    WidgetCheck (N_("Override Fade Out Length"),
+        WidgetBool ("crossfade", "fade_out_override")),
+    WidgetSpin (N_("Fade Out Length:"),
+        WidgetFloat ("crossfade", "fade_out_override_length"),
+        {0.1, 15, 0.1, N_("seconds")},
+        WIDGET_CHILD),
+    WidgetCheck (N_("Use S-Curve for fades"),
+        WidgetBool ("crossfade", "use_sigmoid")),
+    WidgetSpin (N_("S-Curve Steepness:"),
+        WidgetFloat ("crossfade", "sigmoid_steepness"),
+        {2.0, 16, 0.5, N_("(higher is steeper)")},
         WIDGET_CHILD),
     WidgetLabel (N_("<b>Tip</b>")),
     WidgetLabel (N_("For better crossfading, enable\n"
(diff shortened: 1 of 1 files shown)


jlindgren90:
Hi, and first of all, thanks for your contribution! I wrote the current incarnation of the crossfade plugin originally, and I'd be happy to see it add some more DJ-friendly options. I unfortunately don't have time to do a detailed review right now, but would like to do so at some point soon.

You mentioned that the "Override Fade Out Length" option might not actually be needed for your use-case. Would you mind submitting a revision of the pull request without that option added? I want to take some more time to ponder over that one, and the others look simpler to review, so maybe we could get them merged first, and revisit "Override Fade Out Length" in a later pull request if you find you still need/want it.

src/crossfade/crossfade.cc
```diff
@@ -21,6 +21,7 @@
 #include <libaudcore/plugin.h>
 #include <libaudcore/preferences.h>
 #include <libaudcore/runtime.h>
+#include <bits/stdc++.h>
```
jlindgren90, line 24: I think this header is non-standard and is causing the Mac OS build to fail. Is it needed, and is there a standard C++ header you could use instead?

giantg: So this is showing my "greenness" with C/C++ hahaha.  I foolishly included that header just to get ```exp()``` from ```<cmath>```.  Sorry about that.  

After making that swap things still compile here and functionality appears to be identical.  My environment is GCC on ArchLinux (I was going to attempt a Windows cross compile at some point).

giantg:
> Hi, and first of all, thanks for your contribution! I wrote the current incarnation of the crossfade plugin originally, and I'd be happy to see it add some more DJ-friendly options. I unfortunately don't have time to do a detailed review right now, but would like to do so at some point soon.
> 

Thank you for being receptive of someone tinkering in the source!  I get it.  Open source projects are usually more of a "labor of love" vs. a paid product that receives immediate support.  I appreciate your efforts!  Also, if there is anything stylistic regarding the code you want changed you won't hurt my feelings so just let me know.  

> You mentioned that the "Override Fade Out Length" option might not actually be needed for your use-case. Would you mind submitting a revision of the pull request without that option added? I want to take some more time to ponder over that one, and the others look simpler to review, so maybe we could get them merged first, and revisit "Override Fade Out Length" in a later pull request if you find you still need/want it.

I cleaned things up removing those options in the most recent commit.

jlindgren90:
I reworked this a bit and committed it as https://github.com/audacious-media-player/audacious-plugins/commit/18d162aa437f82346451a8af2d631613d1b314a2.

I used `tanh` rather than `exp` to compute the sigmoid as it made the equation slightly nicer (should be equivalent though). I also made the new settings opt-in, so that the defaults are unchanged.

Give it a try and let me know if anything's not working as expected. Thanks again!

PR #144, opened 2023-09-23
Background music effect plugin

emmef:
This is the initial version of the background music plug-in. It dynamically measures "perceptive" loudness (0.4 seconds window) and combines it with a larger, traditional integration time for a nicer listening experience. 
Perceptive loudness is not yet based on the frequency dependent loudness curves.


jlindgren90:
The goal seems to be similar to the existing Dynamic Range Compressor plugin. Can you describe what's different, and why you decided to write a new plugin instead of improving the existing one?

emmef:
I though it would be rash to just change it, as I do not own the compressor code and I do not know if people prefer the sound of the existing compressor. 
Technically, I do not offer clipping protection and the option to reduce the effect of compression with the "dynamic range" control, though that would be trivial to add back again. My approach also eats up more CPU.
However, if you think this is an improvement, I can add the "dynamic range" control and make the target level control in configuration-variable sense compatible with the existing compressor control. I will be able to work on it again in, say, 1 to 2 weeks from now.

jlindgren90:
Can you describe how the effect is different from the existing plugin? For example, is the loudness estimate more accurate? Is the ramp time slower/faster?

Basically, in what way is the existing plugin not suitable for what you're looking for?

emmef:
The product measures the loudness in two ways. The first is slow detection, using an integrated RMS measurement. RMS is already closer to perceived loudness than peak measurement. The second is fast detection, which uses a series of weighed RMS windows from 0.3 to 400 ms, to approach the  IBU R128 0.25 duration/loudness law. The maximum of slow and fast detection is used and smoothed. Prediction is used to turn down peaks naturally and not sounding like someone preventatively turning down the volume knob, like in the current compressor. The slow integration prevents radio-like sound.

**Improvements**
* Perceived loudness also depends on the frequency spectrum of material, so I can add an A-curve, C-curve or the one used by replay gain to detection. Presently a loud kick drum can blot out the rest of the music. But that is also something the current compressor suffers from.
* Doing the processing multi-band (lo and high) is also a possibility. 
* As the plug-in does not prevent clipping, that could also be an addition, though it is likely not a problem below -12 dB target levels because of the way that fast detection works.
* Most values, like the slow integration time, the weight of fast versus slow measurement and the smoothing time are fixed. These can easily be made configurable (within bounds).

jlindgren90:
Thanks for the explanation. Give me some time to think about whether it makes more sense to improve the existing plugin with your new algorithms, or to add a new one. It's been a few years since I last looked at that code.

emmef:
I will be on a conference this whole week, so no hurry!

jlindgren90:
I haven't had time to look at the code in detail, but in principle I'm good with adding this as a separate plugin. Looking at the existing dynamic range plugin, it is very much designed around a particular algorithm, and I don't think there is actually that much of it that could be re-used.

src/background_music/CircularBuffer.h
```diff
@@ -0,0 +1,119 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "utils.h"
+
+template<typename T>
+class CircularBuffer
```
jlindgren90, line 27: We already have a ring buffer container in <libaudcore/ringbuf.h>. Let's not add another.

src/background_music/ConfigValue.h
```diff
@@ -0,0 +1,130 @@
...
+ */
+#include <algorithm>
+#include <cstddef>
+#include <libaudcore/runtime.h>
+#include <limits>
+#include <stdexcept>
+#include <type_traits>
+
+template<typename T>
+struct ConfigValue
```
jlindgren90, line 31: I would prefer not to have this class added, it's a lot of boilerplate what's effectively just a wrapper around the existing libaudcore API. While I appreciate the effort put into designing it, I'd prefer to see the existing API used directly for consistency with the other plugins.

jlindgren90 (changes requested):

emmef:
I removed the wrapper for configuration.
I also realised when rewriting that the circular buffer is a multi-tap delay instead of a ring buffer. 

I _can_ still rewrite it to use the generic RingBuf class, but that would add a lot of clutter: 

- adding, subtraction and remainder operations on top of the ones used in the implementation of RingBuf itself, especially to get samples delayed less than the maximum delay. 
- The result is much less readable code.

jlindgren90:
I think we should improve the existing RingBuf class, rather than duplicating it.

It seems you have basically 3 functions not implemented by the existing RingBuf class:
- `get_oldest_and_set()`
- `peek_back()`
- `fill_with()`

`get_oldest_and_set()` is basically `pop()` + `push()` and doesn't need to exist. However, the existing `RingBuf::pop()` doesn't return the popped value. Let's fix that.

`peek_back()` seems useful to add to `RingBuf()`, but let's call it `nth_from_last()` instead.

`fill_with()` seems useful to add to `RingBuf()`.

emmef:
Thanks. 
Do I wait for these changes to RingBuf on the audacious project?

jlindgren90:
I added the functions you need in audacious master branch.

emmef:
Removed the custom Delay and Replaced it with a RingBuf.

jlindgren90:
Thanks, I'll take a look when I get some time.

src/background_music/DoubleIntegrationTimeRmsDetection.h
```diff
@@ -0,0 +1,143 @@
...
+        update_config();
+        // Configure the integrators
+        release_integration = {SHORT_INTEGRATION, rate};
+        long_integration = {LONG_INTEGRATION / 2.0,
+                            rate}; // Happens before sqrt
+        perceivedLoudness.set_rate_and_value(rate, target_level);
+        read_ahead_ = perceivedLoudness.latency();
+        minimum_detection = target_level / maximum_amplification;
+        double balance = BALANCE;
+        if (balance < 0)
```
jlindgren90, line 73: This condition appears to be always false.

src/background_music/FrameBasedEffectPlugin.h
```diff
@@ -0,0 +1,190 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <libaudcore/plugin.h>
+#include <libaudcore/ringbuf.h>
+
+#include "Detection.h"
+
+template<class D>
+class FrameBasedEffectPlugin : public EffectPlugin
```
jlindgren90, line 29: This does not need to be a template class. Only one class is ever used to satisfy the template parameter D.

src/background_music/FrameBasedEffectPlugin.h
```diff
@@ -0,0 +1,190 @@
...
+};
+
+template<class D>
+bool FrameBasedEffectPlugin<D>::init()
+{
+    if (!detection)
+    {
+        detection = new D;
+    }
+    return after_init();
```
jlindgren90, line 69: Calling into the subclass here (with the virtual `after_init()` function) goes against the usual design pattern and is unnecessary. What you should do instead is add an overriding `init()` function to your subclass (BackgroundMusicPlugin). Make `BackgroundMusicPlugin::init` first call `FrameBasedEffectPlugin::init` and then perform whatever other initialization it needs to do.

src/background_music/FrameBasedEffectPlugin.h
```diff
@@ -0,0 +1,190 @@
...
+    int current_channels = 0, current_rate = 0, channel_last_read = 0;
+    D * detection = nullptr;
+};
+
+template<class D>
+bool FrameBasedEffectPlugin<D>::init()
+{
+    if (!detection)
+    {
+        detection = new D;
```
jlindgren90, line 67: Please `std::unique_ptr` or similar rather than raw `new` + `delete`.

src/background_music/FrameBasedEffectPlugin.h
```diff
@@ -0,0 +1,190 @@
...
+                                                 bool end_of_playlist)
+{
+    return process(data);
+}
+
+template<class D>
+int FrameBasedEffectPlugin<D>::adjust_delay(int delay)
+{
+    auto result = aud::rescale<int64_t>(
+        read_ahead_buffer.len() / current_channels - 1, current_rate, 1000);
```
jlindgren90, line 160: Why minus 1?

src/background_music/FrameBasedEffectPlugin.h
```diff
@@ -0,0 +1,190 @@
...
+    // It is assumed data always contains a multiple of channels, but we don't
+    // care.
+    for (float sample : data)
+    {
+        frame_in[channel_last_read++] = sample;
+        if (channel_last_read == current_channels)
+        {
+            // Processing happens per frame. Because of read-ahead there is not
+            // always output available yet.
+            if (offer_frame_return_if_output())
```
jlindgren90, line 129: So we're performing a multi-level function call (`offer_frame_return_if_output` calls several more functions) on every single frame of audio? As in, 48,000 (or whatever) times per second? This seems quite sub-optimal. Is it really necessary? Is there no way to process a bit more data per function call?

src/background_music/Integrator.h
```diff
@@ -0,0 +1,222 @@
...
+    S input_multiplier_;
+
+public:
+    /**
+     * Below this fraction of a sample the input multiplier cannot be
+     * distinguished from 1.0 and the history multiplier would become de-normal.
+     */
+    static S minimum_samples()
+    {
+        static S min = -1.0 / (std::log(std::numeric_limits<S>::epsilon()));
```
jlindgren90, line 42: Is there a good reason this code needs to depend on the floating-point epsilon? Can you not just choose safe and reasonable constants for `minimum_samples` and `maximum_samples`?

From experience, code that pushes the limits of the floating-point implementation can be extremely hard to debug. I would like to see a really strong justification for why this is necessary.

src/background_music/Integrator.h
```diff
@@ -0,0 +1,222 @@
...
+
+    /**
+     * Calculates the history multiplier from a given input multiplier and vice
+     * versa.
+     * @param multiplier The multiplier
+     * @return The other multiplier
+     */
+    [[maybe_unused]] static inline S calculate_other_multiplier(S multiplier)
+    {
+        return 1 - multiplier;
```
jlindgren90, line 81: We don't need to define a new function for an operation so simple as "one minus x". Doing so makes the code less readable, not more.

src/background_music/Integrator.h
```diff
@@ -0,0 +1,222 @@
...
+        double absolute = fabs(history_multiplier);
+        if (absolute < std::numeric_limits<S>::epsilon()())
+        {
+            return 0;
+        }
+        else if (absolute <= (1 - std::numeric_limits<S>::epsilon()()))
+        {
+            return 1.0 / log(absolute);
+        }
+        throw std::invalid_argument(
```
jlindgren90, line 103: Please fix the design so it does not throw exceptions. The inputs of an effect plugin are well-defined and the processing should be able to produce reasonable output for any input. If nothing else, you can always no-op (output the input data unchanged).

src/background_music/Integrator.h
```diff
@@ -0,0 +1,222 @@
...
+     * result is always positive or zero, but not always a whole number. If the
+     * history multiplier is not valid, std::invalid_argument is thrown.
+     * The choice was made not to return -1 on error, as that easily leads to
+     * segfaults.
+     */
+    [[maybe_unused]] static inline S
+    calculate_samples_from_history_multiplier(S history_multiplier)
+    {
+        double absolute = fabs(history_multiplier);
+        if (absolute < std::numeric_limits<S>::epsilon()())
```
jlindgren90, line 95: Again, please choose suitable constants that avoid pushing the limits of the floating-point implementation.

src/background_music/Integrator.h
```diff
@@ -0,0 +1,222 @@
...
+     * By default, the integrator does not integrate and just passes-through the
+     * input.
+     */
+    inline constexpr Integrator() : history_multiplier_(0), input_multiplier_(1) {}
+
+    /**
+     * Creates an integrator over the provided number of samples.
+     * @param samples The number of samples
+     */
+    [[maybe_unused]] inline explicit Integrator(S samples)
```
jlindgren90, line 118: This is a self-contained plugin, not a general-purpose shared library. If there is unused code, please delete it.

src/background_music/Integrator.h
```diff
@@ -0,0 +1,222 @@
...
+    {
+    }
+
+    /**
+     * Creates an integrator of the the provided number of samples that is
+     * calculated from the time and sample rate.
+     * @param time The number of time units
+     * @param sample_rate The sample-rate samples per unit of time
+     */
+    [[maybe_unused]] inline Integrator(S time, S sample_rate)
```
jlindgren90, line 131: This is a self-contained plugin, not a general-purpose shared library. If there is unused code, please delete it.

src/background_music/Integrator.h
```diff
@@ -0,0 +1,222 @@
...
+        : Integrator(time * sample_rate)
+    {
+    }
+    template<typename T1, typename T2>
+    [[maybe_unused]] inline Integrator(T1 time, T2 sample_rate)
```
jlindgren90, line 136: This is a self-contained plugin, not a general-purpose shared library. If there is unused code, please delete it.

src/background_music/Integrator.h
```diff
@@ -0,0 +1,222 @@
...
+        static_assert(std::is_arithmetic_v<T1>);
+        static_assert(std::is_arithmetic_v<T2>);
+    }
+
+    /**
+     * Integrates value input using history as integrated value.
+     * @param history The historic integrated value
+     * @param input The new input
+     */
+    [[maybe_unused]] inline void integrate(S & history, S input) const
```
jlindgren90, line 149: This is a self-contained plugin, not a general-purpose shared library. If there is unused code, please delete it.

src/background_music/Integrator.h
```diff
@@ -0,0 +1,222 @@
...
+    {
+        history = history_multiplier_ * history + input_multiplier_ * input;
+    }
+
+    /**
+     * Integrates value input using history as integrated value.
+     * @param history The historic integrated value
+     * @param input The new input
+     */
+    [[maybe_unused]] inline S get_integrated(S history, S input) const
```
jlindgren90, line 159: This is a self-contained plugin, not a general-purpose shared library. If there is unused code, please delete it.

src/background_music/Integrator.h
```diff
@@ -0,0 +1,222 @@
...
+    {
+        return history_multiplier_ * history + input_multiplier_ * input;
+    }
+
+
+    /**
+     * Getters
+     */
+    [[nodiscard]] [[maybe_unused]] S history_multiplier() const { return history_multiplier_; }
```
jlindgren90, line 168: This is a self-contained plugin, not a general-purpose shared library. If there is unused code, please delete it.

src/background_music/Integrator.h
```diff
@@ -0,0 +1,222 @@
...
+
+    [[nodiscard]] [[maybe_unused]] S input_multiplier() const { return input_multiplier_; }
```
jlindgren90, line 170: This is a self-contained plugin, not a general-purpose shared library. If there is unused code, please delete it.

src/background_music/Integrator.h
```diff
@@ -0,0 +1,222 @@
...
+
+    /**
+     * Returns the number of samples that is integrated over. This is calculated
+     * from the constants and not totally accurate.
+     */
+    [[nodiscard]] [[maybe_unused]] S estimate_integration_samples() const
```
jlindgren90, line 176: This is a self-contained plugin, not a general-purpose shared library. If there is unused code, please delete it.

src/background_music/Loudness.h
```diff
@@ -0,0 +1,232 @@
...
+            release_ = i;
+        }
+
+        void set_value(uint64_t value)
+        {
+            window_sum_ = static_cast<uint64_t >(std::round(static_cast<double>(value * window_sum_) / window_multiplier_));
+        }
+
+        [[nodiscard]] size_t window_size() const { return window_size_; }
+        [[nodiscard]] [[maybe_unused]] size_t hold_samples() const { return hold_samples_; }
```
jlindgren90, line 147: This is a self-contained plugin, not a general-purpose shared library. If there is unused code, please delete it.

src/background_music/utils.h
```diff
@@ -0,0 +1,72 @@
...
+
+template<typename T>
+static constexpr size_t verified_above_min_below_max(size_t elements,
+                                                     size_t minimum = 1)
+{
+    if (elements > minimum && elements <= get_max_array_size<T>())
+    {
+        return elements;
+    }
+    throw std::invalid_argument(
```
jlindgren90, line 55: Please fix the design so it does not throw exceptions. The inputs of an effect plugin are well-defined and the processing should be able to produce reasonable output for any input. If nothing else, you can always no-op (output the input data unchanged).

src/background_music/utils.h
```diff
@@ -0,0 +1,72 @@
...
+
+template<typename T>
+static T * allocate_if_valid(size_t elements, size_t limit)
+{
+    size_t valid_elements = verified_above_min_below_max<T>(elements, 1);
+    if (limit == 0 || valid_elements <= limit)
+    {
+        return new T[elements];
+    }
+    throw std::invalid_argument("allocate_if_valid(elements,limit): number of "
```
jlindgren90, line 68: Please fix the design so it does not throw exceptions. The inputs of an effect plugin are well-defined and the processing should be able to produce reasonable output for any input. If nothing else, you can always no-op (output the input data unchanged).

src/background_music/Loudness.h
```diff
@@ -0,0 +1,232 @@
...
+    };
+
+    RingBuf<uint64_t> buffer_;
+    WindowedRMS * rms_;
```
jlindgren90, line 151: Please use a suitable container (`Index` or `std::vector` or similar) to manage ownership rather than a raw pointer.

src/background_music/background_music.cc
```diff
@@ -0,0 +1,79 @@
...
+#include "basic_config.h"
+
+/*
+ * The standard "center" volume is set to 0.25 while the dynamic range is set to
+ * zero, meaning all songs will be equally loud according to the detected
+ * loudness, that is.
+ */
+static constexpr const char * const background_music_defaults[] = {
+    CONF_TARGET_LEVEL_VARIABLE, CONF_TARGET_LEVEL_DEFAULT_STRING,
+    CONF_MAX_AMPLIFICATION_VARIABLE, CONF_MAX_AMPLIFICATION_DEFAULT_STRING};
```
jlindgren90, line 36: Missing terminating `NULL`.

src/background_music/basic_config.h
```diff
@@ -0,0 +1,46 @@
+#ifndef AUDACIOUS_PLUGINS_BASIC_CONFIG_H
+#define AUDACIOUS_PLUGINS_BASIC_CONFIG_H
+/*
+ * audacious-plugins/basic_config.h
+ *
+ * Added by michel on 2023-09-23
+ * Copyright (C) 2015-2023 Michel Fleur.
+ * Source https://github.com/emmef/audacious-plugins
```
jlindgren90, line 8: Please remove the links to your personal fork.

src/background_music/utils.h
```diff
@@ -0,0 +1,72 @@
...
+
+template<typename T>
+static inline T decibel_level_to_value(T decibel) {
+    return pow(10.0, 0.05 * decibel);
+}
+
+template<typename T>
+static constexpr size_t get_max_array_size()
+{
+    const size_t max = std::numeric_limits<size_t>::max();
```
jlindgren90, line 42: We should not be allocating arrays anywhere near `std::numeric_limits<size_t>::max()`. Even on a 32-bit system that will be close to 4 GB. An audio player should never consume that much memory.

jlindgren90 (changes requested):

src/background_music/Integrator.h
emmef: In this specific case, integration using a floating point number for the integration constants and integrated values becomes unstable for, say, 80 seconds at 96kHz. Inaccuracy is then about <actual seconds>/80 for smaller values, so 10% for 8 seconds. It is, of course, also possible to just use double, in which case the problem goes practically away.

src/background_music/FrameBasedEffectPlugin.h
emmef: To be honest, I think a lot will be inlined, as the compiler triggers warnings when adding inline keywords. The intention is to move a bit more responsibilities, like the read ahead buffer to the detection class and that might offer some opportunities for optimization.

emmef:
I fixed most or all of the problems you mentioned and also reduced calculation and calls a bit.

src/background_music/LoudnessFrameProcessor.h
```diff
@@ -0,0 +1,189 @@
...
+    RingBuf<float> read_ahead_buffer;
+    int channels_ = 0;
+    int processed_frames = 0;
+
+    static float get_clamped_value(const char * variable, const double minimum,
+                                   const double maximum)
+    {
+        return static_cast<float>(std::clamp(
+            aud_get_double(CONFIG_SECTION_BACKGROUND_MUSIC, variable),
+            static_cast<double>(minimum), static_cast<double>(maximum)));
```
jlindgren90, line 55: `static_cast<double>` is redundant since `minimum` and `maximum` are already `double`

jlindgren90 (approved):
Thanks for addressing all my feedback. I didn't have time to follow through all the math in detail, but the code looks sound.

I had only one trivial comment left. This is fine to merge as far as I'm concerned.

jlindgren90:
Looks like there is a build failure on MacOS - maybe a missing `#include`

src/background_music/LoudnessFrameProcessor.h
emmef: Indeed! I did switch most internal calculations to float, so it was probably a leftover from that.
I also added the header <algorithm> in Integrator.h that defines std::max(a,b)

radioactiveman:
Thanks @emmef, the macOS build is now also green. :+1:

I will review this later or tomorrow, for now I only saw easy to fix style issues. We are getting close to get this merged so your plugin will be included in Audacious 4.4.

Npa972:
@emmef
i use extra stereo, dynamic range compressor, crystalizer and equalizer in the same time and i have a very good sound for my taste. Especially, compressor is very good with classical music.
Can this new plugin "background music effect" be used in conjunction with those I already use?
Or should certain combinations be avoided (compressor etc.) ?
I welcome any advice.
Thanks !

configure.ac
```diff
@@ -68,7 +68,7 @@ dnl ======================
 
 INPUT_PLUGINS="metronom psf tonegen vtx xsf"
 OUTPUT_PLUGINS=""
-EFFECT_PLUGINS="bitcrusher compressor crossfade crystalizer mixer silence-removal stereo_plugin voice_removal echo_plugin"
+EFFECT_PLUGINS="bitcrusher compressor crossfade crystalizer mixer silence-removal stereo_plugin voice_removal echo_plugin background_music"
```
radioactiveman, line 71: Keep this list sorted alphabetically, please also move `echo_plugin` while fixing that.

meson.build
```diff
@@ -244,6 +244,7 @@ if meson.version().version_compare('>= 0.53')
   }, section: 'Other Inputs')
 
   summary({
+    'Background music': true,
```
radioactiveman, line 247: Capitalize "Music" like in configure.ac.

src/background_music/Integrator.h
```diff
@@ -0,0 +1,165 @@
...
+    {
+        return static_cast<T>(integrate(static_cast<double>(input)));
+    }
+
+    void set_output(const double new_value) { integrated_ = new_value; }
+};
+
+/**
+ * Implements an envelope function that follows rises in input - attack -
+ * immediately and on input that is lowe than the current output, optionally the
```
radioactiveman, line 98: Typo: lowe(r)

src/background_music/Loudness.h
```diff
@@ -0,0 +1,194 @@
...
+    };
+
+    static constexpr float get_weight(const float seconds)
+    {
+        const float relative = std::clamp(seconds, perception_peak_seconds,
+                                          perception_center_seconds) /
+                               perception_center_seconds;
+        return powf(relative, perception_weight_power);
+    }
+    static int get_samples(const float seconds, const int rate)
```
radioactiveman, line 53: Missing blank line

src/background_music/background_music.cc
```diff
@@ -0,0 +1,49 @@
...
+               {CONF_TARGET_LEVEL_MIN, CONF_TARGET_LEVEL_MAX, 1.0}),
+    WidgetSpin(N_("Maximum amplification"),
+               WidgetFloat(CONFIG_SECTION_BACKGROUND_MUSIC,
+                           CONF_MAX_AMPLIFICATION_VARIABLE),
+               {CONF_MAX_AMPLIFICATION_MIN, CONF_MAX_AMPLIFICATION_MAX, 1.0}),
+    WidgetLabel(N_("<b>Advanced</b>")),
+    WidgetSpin(
+        N_("Detection Balance\n(-1=perception; 1=slow)"),
+        WidgetFloat(CONFIG_SECTION_BACKGROUND_MUSIC, CONF_BALANCE_VARIABLE),
+        {CONF_BALANCE_MIN, CONF_BALANCE_MAX, 0.1})};
```
radioactiveman, line 37: - Move the unit (like dB) behind the spin button
- Add a colon to the spin button labels
- Please also describe better what the balance actually influences, for example in a separate label.

My suggestion to get the idea:

```diff
diff --git a/src/background_music/background_music.cc b/src/background_music/background_music.cc
index 33b840e4b..8cf0d2b03 100644
--- a/src/background_music/background_music.cc
+++ b/src/background_music/background_music.cc
@@ -22,19 +22,21 @@
 
 static constexpr const PreferencesWidget background_music_widgets[] = {
     WidgetLabel(N_("<b>Background music</b>")),
-    WidgetSpin(N_("Target level (dB)"),
+    WidgetSpin(N_("Target level:"),
                WidgetFloat(CONFIG_SECTION_BACKGROUND_MUSIC,
                            CONF_TARGET_LEVEL_VARIABLE),
-               {CONF_TARGET_LEVEL_MIN, CONF_TARGET_LEVEL_MAX, 1.0}),
-    WidgetSpin(N_("Maximum amplification"),
+               {CONF_TARGET_LEVEL_MIN, CONF_TARGET_LEVEL_MAX, 1.0, N_("dB")}),
+    WidgetSpin(N_("Maximum amplification:"),
                WidgetFloat(CONFIG_SECTION_BACKGROUND_MUSIC,
                            CONF_MAX_AMPLIFICATION_VARIABLE),
                {CONF_MAX_AMPLIFICATION_MIN, CONF_MAX_AMPLIFICATION_MAX, 1.0}),
     WidgetLabel(N_("<b>Advanced</b>")),
     WidgetSpin(
-        N_("Detection Balance\n(-1=perception; 1=slow)"),
+        N_("Detection Balance:"),
         WidgetFloat(CONFIG_SECTION_BACKGROUND_MUSIC, CONF_BALANCE_VARIABLE),
-        {CONF_BALANCE_MIN, CONF_BALANCE_MAX, 0.1})};
+        {CONF_BALANCE_MIN, CONF_BALANCE_MAX, 0.1}),
+    WidgetLabel (N_("<b>Hint</b>")),
+    WidgetLabel(N_("Use a balance of -1 to [...]\nHigher values cause [...]\nFor [...] use a value of 1."))};
 
 static constexpr const PluginPreferences background_music_preferences = {
     {background_music_widgets}};
```

src/background_music/background_music.cc
```diff
@@ -0,0 +1,49 @@
...
+
+static constexpr const PluginPreferences background_music_preferences = {
+    {background_music_widgets}};
+
+static constexpr const char background_music_about[] =
+    N_("Background music (equal loudness) Plugin for Audacious\n"
+       "Copyright 2023 Michel Fleur");
+
+[[maybe_unused]] EXPORT FrameBasedEffectPlugin
+    aud_plugin_instance({N_("Background music (equal loudness)"), PACKAGE,
```
radioactiveman, line 47: I would like to keep plugin names short and precise, without info in parentheses.
Further information belong in the about text, for example like below.

Users and also translators can then better understand what the plugins purpose actually is.

```
Background music Plugin for Audacious\n
Copyright 2023 Michel Fleur\n\n
Allows to get equal loudness for different tracks by [...non-technical description...] (TBD)
```

radioactiveman (changes requested):
cppcheck found also some hints/warnings, some may be false positives though.
I assume at least the unused functions and uninitialized variables should be easy enough to fix.

```
Loudness.h:122:5: warning: Member variable 'PerceptiveRMS::rms_' is not initialized. [uninitMemberVarPrivate]
    WindowedRMS rms_[STEPS + 1];
    ^
Loudness.h:90:9: warning: Member variable 'WindowedRMS::scale_' is not initialized. [uninitMemberVarPrivate]
        float scale_;
        ^
Loudness.h:91:9: warning: Member variable 'WindowedRMS::output_' is not initialized. [uninitMemberVarPrivate]
        float output_;
        ^
LoudnessFrameProcessor.h:69:5: warning: Member variable 'LoudnessFrameProcessor::perceivedLoudness' is not initialized in the constructor. [uninitMemberVar]
    LoudnessFrameProcessor()
    ^
FrameBasedEffectPlugin.h:53:22: style: Parameter 'channels' can be declared as reference to const [constParameterReference]
    void start(int & channels, int & rate) final
                     ^
FrameBasedEffectPlugin.h:53:38: style: Parameter 'rate' can be declared as reference to const [constParameterReference]
    void start(int & channels, int & rate) final
                                     ^
FrameBasedEffectPlugin.h:66:43: style: Parameter 'data' can be declared as reference to const [constParameterReference]
    Index<float> & process(Index<float> & data) final
                                          ^
LoudnessFrameProcessor.h:179:24: style: Consider using std::transform algorithm instead of a raw loop. [useStlAlgorithm]
                sample *= gain;
                       ^
FrameBasedEffectPlugin.h:46:0: style: The function 'cleanup' is never used. [unusedFunction]
    void cleanup() final
^
FrameBasedEffectPlugin.h:101:0: style: The function 'finish' is never used. [unusedFunction]
    Index<float> & finish(Index<float> & data, bool end_of_playlist) final
^
FrameBasedEffectPlugin.h:106:0: style: The function 'adjust_delay' is never used. [unusedFunction]
    int adjust_delay(int delay) final
^
Loudness.h:94:0: style: The function 'get' is never used. [unusedFunction]
        [[nodiscard]] float get() const { return output_; }
^

```

radioactiveman:
And with this plugin enabled, I can reliably crash Audacious by clicking on the time slider or using the arrow keys. Can you reproduce that?
```
audacious: ../src/libaudcore/ringbuf.cc:29: void RingBufBase::get_areas(int, int, Areas&): Assertion `pos >= 0 && len >= 0 && pos + len <= m_len' failed.
```

configure.ac
emmef: Fixed as part of [this commit](https://github.com/audacious-media-player/audacious-plugins/commit/ff69402b28597d41ac73c29e04ec9dbfd79d5a8f)

meson.build
emmef: Fixed as part of [this commit](https://github.com/audacious-media-player/audacious-plugins/commit/ff69402b28597d41ac73c29e04ec9dbfd79d5a8f)

src/background_music/Integrator.h
emmef: Fixed as part of [this commit](https://github.com/audacious-media-player/audacious-plugins/commit/ff69402b28597d41ac73c29e04ec9dbfd79d5a8f)

src/background_music/Loudness.h
emmef: Fixed part of [this commit](https://github.com/audacious-media-player/audacious-plugins/commit/2084df6f3f4bdcc024aa83000de3f1f319ba6d16)

src/background_music/background_music.cc
emmef: Fixed [here](https://github.com/audacious-media-player/audacious-plugins/commit/cc4ae72111bcc55bde3c1eed6121efdafd11a629)

src/background_music/background_music.cc
emmef: Fixed [here](https://github.com/audacious-media-player/audacious-plugins/commit/cc4ae72111bcc55bde3c1eed6121efdafd11a629)

emmef:
I did the following: 
* Pointed in the conversations above where things were fixed
* The crashes when playing with the time slider were [fixed](https://github.com/audacious-media-player/audacious-plugins/commit/42bc3460de8f27e6bcf0fb39b0e366c801e4ff2b). 
* Most of the `cppcheck` things were either fixed by other fixes or false positives that would break the contract of extending the `EffectPlugin` interface.

emmef:
> @emmef i use extra stereo, dynamic range compressor, crystalizer and equalizer in the same time and i have a very good sound for my taste. Especially, compressor is very good with classical music. Can this new plugin "background music effect" be used in conjunction with those I already use? Or should certain combinations be avoided (compressor etc.) ? I welcome any advice. Thanks !

"Background Music" can replace the "compressor" plugin in your setup: it also automates the volume to achieve equal loudness during and between tracks. It uses a more accurate way to measure perceived loudness.

radioactiveman (approved):

radioactiveman:
Thanks for the new plugin @emmef, it is now merged. For further improvements, feel free to submit new pull requests.

emmef:
Just to be clear (I use a different way to collaborate with colleagues):
* I can delete the fork
* For fixes I just pull the main  [master](https://github.com/audacious-media-player/audacious-plugins/) and then issue a pull request

radioactiveman:
Yes, you can delete your fork now (and create a new one based again on master).
Or you keep the fork and rebase/reset it to our master branch.

Npa972:
@emmef Background music is great for working with classical music where levels can vary a lot. Thank you very much !

PR #146, opened 2023-12-17
pipewire: Use decibel-to-linear translation from libaudcore. Closes: #1218

radioactiveman:


radioactiveman:
@jlindgren90: Did I understand/implement this correctly? The initial bug report description confused me because it says we should use a linear scale (which we do already).

And would it be possible to add the helper method to libaudcore for 4.3.2? Or is it not allowed because we have to increase the libversion?

Thanks.

jlindgren90:
Thanks, this looks right to me. A sanity check would be just to listen to it as you increase/decrease the volume: every 10% change should subjectively sound like about the "same" relative volume change. I don't have pipewire installed, so if you're happy with how it sounds, please go ahead and merge.

It occurs to me that it may be possible to make use of the existing translation in libaudcore via the `audio_amplify` function, without adding a new helper. I think this should work and be equivalent to what you have currently (not tested):

```
diff --git a/src/pipewire/pipewire.cc b/src/pipewire/pipewire.cc
index c2838359e..1a0159502 100644
--- a/src/pipewire/pipewire.cc
+++ b/src/pipewire/pipewire.cc
@@ -138,16 +138,13 @@ void PipeWireOutput::set_volume(StereoVolume v)
 
     float values[m_channels];
 
-    if (m_channels == 2)
-    {
-        values[0] = v.left / 100.0f;
-        values[1] = v.right / 100.0f;
-    }
-    else
-    {
-        for (unsigned int i = 0; i < m_channels; i++)
-            values[i] = aud::max(v.left, v.right) / 100.0f;
-    }
+    // Re-use libaudcore's decibel-to-linear translation by passing
+    // full-scale PCM values (1.0) to audio_amplify(). This also
+    // translates the StereoVolume to an arbitrary number of channels.
+    for (unsigned int i = 0; i < m_channels; i++)
+        values[i] = 1.0f;
+
+    audio_amplify(values, m_channels, 1, v);
 
     pw_thread_loop_lock(m_loop);
     pw_stream_set_control(m_stream, SPA_PROP_channelVolumes, m_channels, values, nullptr);
```

jlindgren90:
LGTM

radioactiveman:
@jlindgren90: Your suggestion works as expected, thanks. :+1:

PR #147, opened 2023-12-29
statusicon-qt: Set proper tooltip

radioactiveman:
Even though we override the QSystemTrayIcon::event() method,
it is never called by Qt (at least on Linux when the panel/DE
supports StatusNotifierItem). Therefore our custom info popup
is not shown. Set a proper tooltip as workaround.

### DIFF
--- a/src/statusicon-qt/statusicon.cc
+++ b/src/statusicon-qt/statusicon.cc
@@ -17,13 +17,13 @@
  * the use of this software.
  */
 
-#include <libaudcore/i18n.h>
 #include <libaudcore/drct.h>
 #include <libaudcore/hook.h>
-#include <libaudcore/plugin.h>
-#include <libaudcore/runtime.h>
+#include <libaudcore/i18n.h>
 #include <libaudcore/interface.h>
+#include <libaudcore/plugin.h>
 #include <libaudcore/preferences.h>
+#include <libaudcore/runtime.h>
 
 #include <libaudqt/menu.h>
 
@@ -32,7 +32,8 @@
 #include <QSystemTrayIcon>
 #include <QWheelEvent>
 
-class StatusIcon : public GeneralPlugin {
+class StatusIcon : public GeneralPlugin
+{
 public:
     static const char about[];
     static const char * const defaults[];
@@ -53,6 +54,8 @@ public:
     bool init ();
     void cleanup ();
 
+private:
+    static void update_tooltip (void * data, void * user_data);
     static void window_closed (void * data, void * user_data);
     static void activate (QSystemTrayIcon::ActivationReason);
     static void open_files ();
@@ -161,7 +164,10 @@ bool SystemTrayIcon::event (QEvent * e)
     {
     case QEvent::ToolTip:
         if (! aud_get_bool ("statusicon", "disable_popup"))
+        {
+            setToolTip (QString ()); /* prevent double tooltip */
             audqt::infopopup_show_current ();
+        }
         return true;
 
     case QEvent::Wheel:
@@ -188,6 +194,11 @@ bool StatusIcon::init ()
     tray->setContextMenu (menu);
     tray->show ();
 
+    update_tooltip (nullptr, nullptr);
+
+    hook_associate ("title change", update_tooltip, nullptr);
+    hook_associate ("playback ready", update_tooltip, nullptr);
+    hook_associate ("playback stop", update_tooltip, nullptr);
     hook_associate ("window close", window_closed, nullptr);
 
     return true;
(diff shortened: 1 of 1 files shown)


radioactiveman:
@jlindgren90: Why is the `title_changed()` method never called? What am I missing here? It is registered to the hook and there is no typo.

jlindgren90:
> @jlindgren90: Why is the `title_changed()` method never called? What am I missing here? It is registered to the hook and there is no typo.

"title change" is for streaming when the title changes mid-stream. To get the initial title you can use "playback ready" (if memory serves).

radioactiveman:
Thanks, updated accordingly.

One more question: I wonder if there is a difference regarding number of memory allocations and preventing memory leaks between these options. Which would you choose?

```c++
const char * title = aud_drct_get_title ();
tray->setToolTip (title);
```

```c++
String title = aud_drct_get_title ();
tray->setToolTip ((const char *) title);
```

```c++
String title = aud_drct_get_title ();
tray->setToolTip (QString (title));
```

jlindgren90:
> One more question: I wonder if there is a difference regarding number of memory allocations and preventing memory leaks between these options. Which would you choose?
> 
> ```c++
> const char * title = aud_drct_get_title ();
> tray->setToolTip (title);
> ```
> 
> ```c++
> String title = aud_drct_get_title ();
> tray->setToolTip ((const char *) title);
> ```
> 
> ```c++
> String title = aud_drct_get_title ();
> tray->setToolTip (QString (title));
> ```

The first one is unsafe since the String will go out of scope too early (leaving a dangling pointer). The last two are equivalent so either one is fine.

jlindgren90:
Thanks for working on this. Have you checked how this works with a panel/desktop environment supporting only XEmbed? Do we end up showing a double tooltip/info popup in that case?

radioactiveman:
> Thanks for working on this. Have you checked how this works with a panel/desktop environment supporting only XEmbed? Do we end up showing a double tooltip/info popup in that case?

I have tested it with Xfce and GNOME. Xfce has shown a tooltip also before ("Audacious") which is rather useless. GNOME with this [extension](https://extensions.gnome.org/extension/615/appindicator-support/) neither shows a tooltip nor the info popup. Next week I can test how macOS behaves.

Which panel / DE do you use? Could you please test it yourself too?

radioactiveman:
@bpanca05: In your [bug report](https://github.com/orgs/audacious-media-player/discussions/93) I have seen you use KDE and also enabled the status icon plugin. Could you please answer my questions and try the patch from this pull request?

- Does KDE show the info popup like in my screenshot below, a general tooltip like "Audacious" or no tooltip at all?
- Does changing the volume work with KDE when using the scroll wheel above the icon?
- Does KDE show both the info popup and a tooltip with the patch applied (worse behavior than before)?

Thanks for feedback.

info-popup

jlindgren90:
To test XEmbed, try `lxpanel` (the old GTK version, not lxqt-panel). Here it shows a double tooltip:
image

jlindgren90:
This seems to fix the double-tooltip issue:
```
diff --git a/src/statusicon-qt/statusicon.cc b/src/statusicon-qt/statusicon.cc
index e22bbe143..c976b3bde 100644
--- a/src/statusicon-qt/statusicon.cc
+++ b/src/statusicon-qt/statusicon.cc
@@ -164,7 +164,10 @@ bool SystemTrayIcon::event (QEvent * e)
     {
     case QEvent::ToolTip:
         if (! aud_get_bool ("statusicon", "disable_popup"))
+        {
+            setToolTip(QString()); /* prevent double tooltip */
             audqt::infopopup_show_current ();
+        }
         return true;
 
     case QEvent::Wheel:
```

radioactiveman:
Thanks John for testing and the double tooltip fix. Meanwhile I have tried it on macOS where the tooltip is shown now.

Happy New Year! :fireworks:

bpanca05:
> @bpanca05: In your [bug report](https://github.com/orgs/audacious-media-player/discussions/93) I have seen you use KDE and also enabled the status icon plugin. Could you please answer my questions and try the patch from this pull request?
> 
>     * Does KDE show the info popup like in my screenshot below, a general tooltip like "Audacious" or no tooltip at all?
> 
>     * Does changing the volume work with KDE when using the scroll wheel above the icon?
> 
>     * Does KDE show both the info popup and a tooltip with the patch applied (worse behavior than before)?
> 
> 
> Thanks for feedback.
> 
> info-popup

Sorry for the late response, but before and after this commit merged, nothing change from status icon on KDE, it's simply say "Audacious" and no additional information like your attached image.

image

PR #150, opened 2024-01-31
aac: Avoid crash with corrupt files

radioactiveman:
NeAACDecInit() returns -1 in case of an error.
Subtracting -1 actually increases the buflen by 1,
resulting in a buffer overflow in memmove().

See also: https://github.com/orgs/audacious-media-player/discussions/109


src/aac/aac.cc
```diff
@@ -319,7 +319,13 @@ static void aac_seek (VFSFile & file, NeAACDecHandle dec, int time, int len,
     unsigned char chan;
     unsigned long rate;
 
-    if ((used = NeAACDecInit (dec, (unsigned char *) buf, * buflen, & rate, & chan)))
+    if ((used = NeAACDecInit (dec, (unsigned char *) buf, * buflen, & rate, & chan)) < 0)
+    {
+        AUDERR ("Failed to initialize AAC decoder.\n");
+        return;
```
jlindgren90, line 325: If `NeAACDecInit()` returns an error, then is it still safe to proceed with calling `NeAACDecDecode()`? If not, perhaps we should make `aac_seek()` return an error in that case so that `AACDecoder::play()` can break out of the decode loop.

jlindgren90 (commented):
Thanks. Looks good to me except for one concern regarding `aac_seek()`.

src/aac/aac.cc
radioactiveman: Good catch. Can we use `* buflen = 0;` like a few lines above when checking the result of `aac_probe()`?

jlindgren90: Looks like that should work, yes.

PR #151, opened 2024-02-01
lyrics: Restore GTK support

radioactiveman:
- Make it possible to reuse most code with the Qt plugin
- Rename to lyrics since lyricwiki is no longer used as source


radioactiveman:
@jlindgren90: This is still a draft, but I would like to get some feedback and also help regarding Autotools.

It is possible to use a single Makefile and directory? My current approach with referencing relative paths is not ideal. Something like this does not work unfortunately.

```Makefile
SRCS = chart_lyrics_provider.cc \
       file_provider.cc \
       lyrics_ovh_provider.cc \
       utils.cc

include ../../buildsys.mk
include ../../extra.mk

plugindir := ${plugindir}/${GENERAL_PLUGIN_DIR}

LD = ${CXX}

CFLAGS += ${PLUGIN_CFLAGS}

ifeq ($(USE_GTK),yes)
PLUGIN = lyrics${PLUGIN_SUFFIX}
SRCS += lyrics.cc
CPPFLAGS += ${PLUGIN_CPPFLAGS} ${GTK_CFLAGS} ${GLIB_CFLAGS} ${JSON_GLIB_CFLAGS} ${XML_CFLAGS} -I../..
LIBS += ${GTK_LIBS} ${GLIB_LIBS} ${JSON_GLIB_LIBS} ${XML_LIBS} -laudgui
endif

ifeq ($(USE_QT),yes)
PLUGIN = lyrics-qt${PLUGIN_SUFFIX}
SRCS += lyrics-qt.cc
CPPFLAGS += ${PLUGIN_CPPFLAGS} ${QT_CFLAGS} ${GLIB_CFLAGS} ${XML_CFLAGS} -I../..
LIBS += ${QT_LIBS} ${GLIB_LIBS} ${XML_LIBS}
endif
```

src/lyrics/lyrics-qt.cc
```diff
@@ -0,0 +1,304 @@
...
+        WidgetBool (CFG_SECTION, "use-embedded")),
+    WidgetCombo (N_("Fetch lyrics from internet:"),
+        WidgetString (CFG_SECTION, "remote-source"),
+        {{remote_sources}}),
+    WidgetCheck (N_("Store fetched lyrics in local cache"),
+        WidgetBool (CFG_SECTION, "enable-cache")),
+    WidgetLabel (N_("<b>Local Storage</b>")),
+    WidgetCheck (N_("Load lyric files (.lrc) from local storage"),
+        WidgetBool (CFG_SECTION, "enable-file-provider"))
+};
```
radioactiveman, line 97: Could `defaults` and `widgets` be also shared between Qt and GTK somehow?

src/lyrics/lyrics-qt.cc
```diff
@@ -0,0 +1,304 @@
...
+    cursor.insertText (lyrics);
+}
+
+bool try_parse_json (const Index<char> & buf, const char * key, String & output)
+{
+    QByteArray json = QByteArray (buf.begin (), buf.len ());
+    QJsonDocument doc = QJsonDocument::fromJson (json);
+
+    if (doc.isNull () || ! doc.isObject ())
+        return false;
```
radioactiveman, line 138: Should we call `output = String();` in error cases explicitly?

src/lyrics/lyrics.cc
```diff
@@ -0,0 +1,330 @@
...
+        WidgetBool (CFG_SECTION, "use-embedded")),
+    WidgetCombo (N_("Fetch lyrics from internet:"),
+        WidgetString (CFG_SECTION, "remote-source"),
+        {{remote_sources}}),
+    WidgetCheck (N_("Store fetched lyrics in local cache"),
+        WidgetBool (CFG_SECTION, "enable-cache")),
+    WidgetLabel (N_("<b>Local Storage</b>")),
+    WidgetCheck (N_("Load lyric files (.lrc) from local storage"),
+        WidgetBool (CFG_SECTION, "enable-file-provider"))
+};
```
radioactiveman, line 83: libaudgui does not show all widgets for some reason. This needs further verification.

src/lyrics/lyrics.cc
```diff
@@ -0,0 +1,330 @@
...
+    Tuple tuple = aud_drct_get_tuple ();
+    g_state.title = tuple.get_str (Tuple::Title);
+    g_state.artist = tuple.get_str (Tuple::Artist);
+    g_state.lyrics = String ();
+
+    if (aud_get_bool (CFG_SECTION, "split-title-on-chars"))
+    {
+        // TODO: Implement this also for GTK with g_regex
+        AUDINFO ("Regex support is not yet implemented in the GTK Lyrics Plugin\n");
+    }
```
radioactiveman, line 168: Regex support is not yet implemented

src/lyrics/lyrics.cc
```diff
@@ -0,0 +1,330 @@
...
+    {
+        if (remote_provider)
+        {
+            String edit_uri = remote_provider->edit_uri (g_state);
+
+            // TODO: Is passing 'edit_uri' to 'edit_lyrics_cb' like this safe?
+            if (edit_uri && edit_uri[0])
+                append_item_to_menu (menu, _("Edit Lyrics ..."),
+                 (GCallback) edit_lyrics_cb, aud::to_ptr ((const char *) edit_uri));
+        }
```
radioactiveman, line 276: How would you pass `edit_uri` to the callback function?

src/lyrics/lyrics.h
```diff
@@ -0,0 +1,142 @@
...
+struct LyricsState {
+    String filename; // of song file
+    String title, artist;
+    String lyrics;
+
+    enum Source {
+        None,
+        Embedded,
+        Local,
+        LyricWiki, // TODO: Can this be safely removed?
```
radioactiveman, line 43: It is no longer used, but affects the value of entries below. Can we remove it anyway?

jlindgren90:
Thanks. I'll take a look at this when I can and try to answer the specific questions. I know in advance it will be good quality, your work always is :)

src/lyrics/lyrics.cc
```diff
@@ -0,0 +1,375 @@
...
+                "^.*\\s+[", aud_get_str (CFG_SECTION, "truncate-on-chars"), "]\\s+"
+            });
+
+            StringBuf title_pattern = str_concat ({
+               "\\s+[", aud_get_str (CFG_SECTION, "truncate-on-chars"), "]\\s+.*$"
+            });
+
+            // TODO: Are these reassignments safe and not leaking memory?
+            artist = CharPtr (truncate_by_pattern (artist, artist_pattern));
+            title = CharPtr (truncate_by_pattern (title, title_pattern));
```
radioactiveman, line 188: I assume the results of `g_match_info_fetch()` are leaked here. Is this correct? If yes, what's the cleanest way to free and reassign the strings?

radioactiveman:
> Thanks. I'll take a look at this when I can and try to answer the specific questions. I know in advance it will be good quality, your work always is :)

Thanks John. :) It's sufficient if you review the GTK plugin and answer my questions. All other code was extracted from the Qt plugin and is unchanged.

When I know how to structure the directory/files, I will split the changes into multiple commits (prepare code sharing + restore GTK plugin).

src/lyrics/lyrics.h
jlindgren90: I think we can remove `LyricWiki`. It doesn't look like `enum Source` is used directly in the config file.

src/lyrics/lyrics-qt.cc
jlindgren90: Sure, you could put them in `preferences.h`. They can be `static` and don't need to be members of `Lyrics` or `LyricsQt`.

src/lyrics/lyrics-qt.cc
jlindgren90: I don't think that's necessary. It's fine to leave `output` unchanged in error cases.

src/lyrics/lyrics.cc
jlindgren90: It looks like libaudgui does not support nested `WidgetTable`. See `fill_table()` in prefs-widget.cc.

src/lyrics/lyrics.cc
jlindgren90: I would make a copy of the string with `g_strdup()`, then instead of `g_signal_connect()` use `g_signal_connect_data()` and pass `g_free()` as the `destroy_data` parameter.

https://docs.gtk.org/gobject/func.signal_connect_data.html

src/lyrics/lyrics.cc
jlindgren90: This is actually fine, nothing is leaked. `CharPtr` has overloaded `operator=` which handles freeing the previous pointer.

jlindgren90:
For directory structure, what you have already is okay, but my suggestion is to have 3 folders:
- `src/lyrics-common` contains the toolkit-agnostic source files (but no Makefile etc.)
- `src/lyrics-gtk` contains the GTK-specific source file (`lyrics-gtk.cc`) and the Makefile etc. for the GTK plugin
- `src/lyrics-qt` contains the Qt-specific source file (`lyrics-qt.cc`) and the Makefile etc. for the Qt plugin

It's fine for the Makefiles to use relative paths to access the source files in `src/lyrics-common`. This is what we've done already with e.g. the `src/ui-common` folder.

jlindgren90:
I guess an advantage of your current approach is that there's just a single `meson.build`. I unfortunately don't know an easy way to do anything similar (i.e. use a single folder and Makefile) with the `buildsys.mk` system. It seems to assume one Makefile per plugin.

I think I'd slightly prefer to split the `meson.build` and use 3 folders as I described above. It's a little more hassle but I think it's less confusing in the end. Your call.

radioactiveman:
Thanks again @jlindgren90 for the review and suggestions. I went ahead and merged it myself with the suggested changes.

If you have an idea how to make autotools work on mingw64 (see pull request https://github.com/audacious-media-player/audacious-plugins/pull/152), please let me know.
For now I have added a workaround: https://github.com/audacious-media-player/audacious-plugins/commit/98be4f10679c5a1d931474040c3638f9d09a82b3

PR #152, opened 2024-02-12
CI: Use 64-bit for Windows and disable autotools build

radioactiveman:
See also: https://www.msys2.org/news/#2023-12-13-starting-to-drop-some-32-bit-packages


jlindgren90:
I haven't tried a 64-bit Windows build locally yet (our last official Windows build is 32-bit). Perhaps the next one should be 64-bit since upstream 32-bit support is being phased out.

Somewhat related: our download page states that we support Windows 7 (and I think our last build did in fact run on Window 7). Since upstream Windows 7 (& 8) support is also being phased out, we probably need to update that for the next release as well.

PR #153, opened 2024-03-02
lyrics: Remove French prefix which lyrics.ovh embeds into lyrics

radioactiveman:
https://lyrics.ovh embeds a French prefix for `Lyrics of the song {title} by {artist}` to all lyrics,
regardless of the lyrics language or user country (verified with proxies).

Examples:
- https://api.lyrics.ovh/v1/R.E.M./Shiny%20Happy%20People
- https://api.lyrics.ovh/v1/R.E.M./At%20My%20Most%20Beautiful

Questions:
- Is it safe to return the char pointer like this or should it be wrapped into a String or StringBuf?
- Is there a way to trim also non-breaking-space characters? `g_strchug()` does not remove them.
The 2nd example shows therefore unnecessary newlines below the header.

grafik

### DIFF
--- a/src/lyrics-common/lyrics_ovh_provider.cc
+++ b/src/lyrics-common/lyrics_ovh_provider.cc
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2010-2019 Ariadne Conill <ariadne@dereferenced.org>
+ * Copyright (c) 2024 Thomas Lange <thomas-lange2@gmx.de>
  *
  * Permission to use, copy, modify, and/or distribute this software for any
  * purpose with or without fee is hereby granted, provided that the above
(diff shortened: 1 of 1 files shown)


jlindgren90:
>     * Is it safe to return the char pointer like this or should it be wrapped into a String or StringBuf?

No, it's not safe because the char pointer points inside the temporary `StringBuf`.

>     * Is there a way to trim also non-breaking-space characters? `g_strchug()` does not remove them.
>       The 2nd example shows therefore unnecessary newlines below the header.

`g_strchug()` handles ASCII only, but GLib has some UTF-8 functions that should work for this. I think this should work:

```
static String truncate_prefix (const String & lyrics, const char * prefix)
{
    if (! str_has_prefix_nocase (lyrics, prefix))
        return lyrics;

    const char * truncated = strstr (lyrics, "\r\n");
    if (! truncated || ! g_utf8_validate (truncated, -1, NULL))
        return lyrics;

    while ((* truncated) && g_unichar_isspace (g_utf8_get_char (truncated)))
        truncated = g_utf8_next_char (truncated);

    return String (truncated);
}
```

radioactiveman:
Works perfectly, thanks John.

PR #155, opened 2024-03-17
scrobbler2: Make plugin work on macOS (and potentially Windows too)

radioactiveman:
The GLib mainloop and timers don't run here when using Qt.
Make use of QueuedFunc from libaudcore to fix this.

See also: https://github.com/orgs/audacious-media-player/discussions/111

### DIFF
--- a/src/scrobbler2/config_window.cc
+++ b/src/scrobbler2/config_window.cc
@@ -20,11 +20,13 @@ gboolean          invalidate_session_requested = false;
 enum permission perm_result                  = PERMISSION_UNKNOWN;
 String          username;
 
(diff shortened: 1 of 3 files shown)


radioactiveman:
@jlindgren90: Do you have a Last.fm account to test this? I could only verify that the permission check works now on macOS (meaning the dialog with the permission URL is shown).

jlindgren90:
I have a last.fm account and will try this out when I get a chance -- can't promise when exactly. Thanks.

jlindgren90:
Seems to work! Let's go for it.

PR #157, opened 2024-04-06
blur_scope-qt: fix incorrect glib.h include path

jkl1337:
The include path given as glib-2.0/glib.h is not correct for meson/pkg-config. This usually works on systems where glib-2.0 is in a system include path, but will break when this is not the case. Changed to match other modules.

### DIFF
--- a/src/blur_scope-qt/blur_scope.cc
+++ b/src/blur_scope-qt/blur_scope.cc
@@ -26,7 +26,7 @@
 
 #include <math.h>
 #include <string.h>
-#include <glib-2.0/glib.h>
+#include <glib.h>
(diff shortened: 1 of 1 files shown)


jlindgren90:
Thanks!

radioactiveman:
@jkl1337: Out of interest, on which system / distribution has this caused problems?

jkl1337:
@radioactiveman 

I just happened to catch it on a MacOS system where glib-2.0 was installed by homebrew but not symlinked (brew linked) into the system include path. But I verified on multiple systems that glib pkg-config assumes the fixed include style.

PR #158, opened 2024-04-07
Made titlebar reflect focus status

y0shir:
Made titlebar of main window when using classic skins reflect focus status.

This was bugging me, so I decided to fix it, hopefully I did it on acceptable level.

PR #159, opened 2024-04-11
skins: Made titlebar reflect focus status when using classic skins in gtk mode

y0shir:
This is a rework of merge request [#158 ](y0shir:master)
This makes main, equalizer and playlist window all use focused versions of titlebars when focus in on either of them and unfocused when none of them have focus.

### DIFF
--- a/src/skins/dock.cc
+++ b/src/skins/dock.cc
@@ -373,3 +373,8 @@ bool dock_is_focused()
     }
     return false;
 }
+
+void refresh_dock_focus(){
+	for (DockWindow & dw : windows)
+        { dw.window -> focus_refresh (); }
+}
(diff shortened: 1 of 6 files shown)


radioactiveman:
> This is my first time contributing to open source project, so I am not sure if I didn't do something. Should have I updated copyright?

Sorry for the missing / late reply. My focus right now is to release Audacious 4.4 soon. And thanks for your contribution. :+1:

What I noticed so far is:
- Implement the feature like it was done for skins-qt (see my commit in https://github.com/audacious-media-player/audacious-plugins/pull/158). The setting `Display active titlebar when any window is focused` is for example missing here.
- The code style is wrong in many aspects (tabs instead of spaces, placement of parentheses and curly brackets, ...). Please use the same style as in the existing code.

jlindgren90:
I implemented this a bit differently in 8c2e98333cf74b6fc49288644d8815ecc1551565, following the Qt version. Thanks.

PR #161, opened 2024-06-02
Wayland fixes

radioactiveman:
Fixes for Audacious when running on Wayland, see the commit messages for more details.

Tested with `QT_QPA_PLATFORM=wayland audacious` and `GDK_BACKEND=wayland audacious --gtk` on Ubuntu 24.04.

### DIFF
--- a/src/gtkui/ui_gtk.cc
+++ b/src/gtkui/ui_gtk.cc
@@ -980,7 +980,6 @@ bool GtkUI::init ()
 
     timer_add (TimerRate::Hz4, ui_volume_slider_update, volume);
 
-    g_signal_connect (window, "map-event", (GCallback) pl_notebook_grab_focus, nullptr);
     g_signal_connect (window, "delete-event", (GCallback) window_delete, nullptr);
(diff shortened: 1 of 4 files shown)


radioactiveman:
Another option for the Winamp interface would be to print a warning only instead of bailing out.
I prefer do do the latter, because it's simply not usable on Wayland.

src/gtkui/ui_gtk.cc
```diff
@@ -980,7 +980,6 @@ bool GtkUI::init ()
 
     timer_add (TimerRate::Hz4, ui_volume_slider_update, volume);
 
-    g_signal_connect (window, "map-event", (GCallback) pl_notebook_grab_focus, nullptr);
```
jlindgren90, line 983: Does the playlist still get focused at startup?

jlindgren90: Ah I see you addressed this in the commit message.

jlindgren90 (approved):
LGTM

PR #162, opened 2024-06-05
Set parent window for infopopup

radioactiveman:
Follow-up for the infopopup changes in libaudgui from https://github.com/audacious-media-player/audacious/pull/1399.

### DIFF
--- a/src/gtkui/gtkui.h
+++ b/src/gtkui/gtkui.h
@@ -37,6 +37,7 @@ extern Playlist menu_tab_playlist;
 extern const PluginPreferences gtkui_prefs;
 
 /* ui_gtk.c */
+GtkWindow * get_main_window ();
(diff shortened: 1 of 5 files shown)


src/gtkui/gtkui.h
```diff
@@ -37,6 +37,7 @@ extern Playlist menu_tab_playlist;
 extern const PluginPreferences gtkui_prefs;
 
 /* ui_gtk.c */
+GtkWindow * get_main_window ();
```
radioactiveman, line 40: What would you prefer? Getter function like this or `extern GtkWidget * window;` in `gtkui.h`?

jlindgren90: This is fine :)

jlindgren90:
Looks good. Thanks!

PR #167, opened 2024-08-09
Lower min speed to same as YouTube. Closes: #1447

petarturukalo:
See https://github.com/audacious-media-player/audacious/issues/1447.

image


radioactiveman:
Merged, thanks for the contribution and testing.

PR #168, opened 2024-08-14
CI: Switch from MINGW64 to UCRT64

radioactiveman:
UCRT64 is meanwhile the recommended MSYS2 environment for Windows.
Visual Studio uses it by default since 2015.

See also:
- https://www.msys2.org/docs/environments/
- https://www.msys2.org/news/#2022-10-29-changing-the-default-environment-from-mingw64-to-ucrt6


jlindgren90:
Thanks!

PR #170, opened 2024-09-14
qtaudio: Add Qt 6 support

radioactiveman:
Draft status, still missing:
- [periodSize()](https://doc.qt.io/qt-5/qabstractaudiooutput.html#periodSize) is not available in Qt 6. What can we use instead?
- How do we need to convert the audio data for signed/unsigned formats and endianess?

[Qmmp](https://sourceforge.net/p/qmmp-dev/code/HEAD/tree/trunk/qmmp/src/plugins/Output/qtmultimedia/outputqtmultimedia.cpp) uses:
```c
#if Q_BYTE_ORDER == Q_BIG_ENDIAN
    if(format == Qmmp::PCM_S16LE)
        format = Qmmp::PCM_S16BE;
    else if(format == Qmmp::PCM_S32LE)
        format = Qmmp::PCM_S32BE;
#endif
```

From https://doc.qt.io/qt-6/qaudioformat.html#SampleFormat-enum:

> Qt will always expect and use samples in the endianness of the host platform. When processing audio data from external sources yourself, ensure you convert them to the correct endianness before writing them to a [QAudioSink](https://doc.qt.io/qt-6/qaudiosink.html) or [QAudioBuffer](https://doc.qt.io/qt-6/qaudiobuffer.html).

@jlindgren90, @kaniini: Could you please support here?


jlindgren90:
> How do we need to convert the audio data for signed/unsigned formats and endianess?

I don't think the plugin needs to do any conversions. Just make open_audio return false for the non-native (byte-swapped) formats. Audacious core already tries to use the native formats IIRC.

Not sure about the period size thing. Generally I don't think Audacious core cares about the period size as long as period_wait works.

src/qtaudio/qtaudio.cc
```diff
@@ -89,18 +97,36 @@ static bool paused;
...
     enum QAudioFormat::SampleType sample_type;
-    enum QAudioFormat::Endian endian;    
+    enum QAudioFormat::Endian endian;
+#endif
 };
 
 static constexpr FormatDescriptionMap FormatMap[] = {
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+    {FMT_S16_LE, QAudioFormat::Int16},
+    {FMT_S16_BE, QAudioFormat::Int16},
```
jlindgren90, line 121: I think we want just the `_NE` formats here.

src/qtaudio/qtaudio.cc
```diff
@@ -205,29 +243,33 @@ void QtAudio::close_audio ()
...
 
-void QtAudio::period_wait ()
+void QtAudioOutput::period_wait ()
 {
     pthread_mutex_lock (& mutex);
 
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+    /* periodSize() is Qt 5 only, how to replace it? */
+#else
     while (output_instance->bytesFree () < output_instance->periodSize ())
```
jlindgren90, line 253: `period_wait` just needs to wait until `write_audio` will write *some* data (more than 0 bytes). So I think this can be as simple as:

```
while (output_instance->bytesFree () == 0)
```

It would be good to check that CPU usage doesn't spike to 100% (one core) or anything.

radioactiveman: The CPU usage is then still way too high. I assume because `pthread_cond_timedwait()` is used incorrectly. It expects an absolute time, not a relative one like in `static const timespec fifty_ms = {0, 50000000}`.

ALSA uses `clock_gettime()` instead. Is this approach fine to use in `period_wait()` and also `drain()`?

https://github.com/audacious-media-player/audacious-plugins/blob/ba579461cbec865c225e9ca1a32cf6f41b3fb5ad/src/alsa/alsa.cc#L466-L485

src/qtaudio/qtaudio.cc
radioactiveman: You probably meant `_LE`?

jlindgren90: No - NE. Native-endian. It is an alias for either LE or BE depending on the target.

src/qtaudio/qtaudio.cc
jlindgren90: Something like that should work, yes

radioactiveman: I have added a separate commit to fix the excessive CPU usage which also occurred with Qt 5.

A simpler fix would be to use `nanosleep()` but within pthread mutexes using `pthread_cond_timedwait()` should be better, right?

src/qtaudio/qtaudio.cc
radioactiveman: So `FormatMap[]` should include for Big Endian all BE (+ FLOAT) formats and for Little Endian all LE (+ FLOAT)? Otherwise explain it with a code example please.

jlindgren90: Just include (for example):

```
{FMT_S16_NE, QAudioFormat::Int16},
```

Omit the explicit LE and BE formats.

src/qtaudio/qtaudio.cc
jlindgren90: Looks good. The advantage of pthread_timed_wait() is that it can return immediately if flush() is called. Just sleeping for 50ms could make seeking/stopping laggy.

src/qtaudio/qtaudio.cc
radioactiveman: Thanks, got it now. Sometimes you think too complicated. :smiley:
For Qt 5 I have kept both signed and unsigned formats. Qt 6 supports only signed ones.

jlindgren90:
Looks good. Thanks!

PR #171, opened 2024-10-16
gtkui: Hide info area text containing emoji properly. Closes: #1491

radioactiveman:
Fading out text with emoji characters can lead to them still being displayed above the text of a new track.
Ignoring zero alpha channels avoids this.

Reproduced on Ubuntu 24.04. Arch Linux requires "noto-fonts-emoji".

### DIFF
--- a/src/gtkui/ui_infoarea.cc
+++ b/src/gtkui/ui_infoarea.cc
@@ -30,6 +30,9 @@
 
 #include "ui_infoarea.h"
 
+#define ALPHA_STEPS 10
+static inline float TO_ALPHA (int steps) { return (float) steps / ALPHA_STEPS; }
+
 #define VIS_BANDS 12
 #define VIS_DELAY 2 /* delay before falloff in frames */
 #define VIS_FALLOFF 2 /* falloff in decibels per frame */
@@ -56,7 +59,7 @@ typedef struct {
     String title, artist, album;
     String last_title, last_artist, last_album;
     AudguiPixbuf pb, last_pb;
-    float alpha, last_alpha;
+    int alpha_steps, last_alpha_steps;
 
     bool show_art;
     bool stopped;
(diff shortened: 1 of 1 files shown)


radioactiveman:
@jlindgren90: Do you think this is a valid fix? I'm still not sure how and why it works. My first guess was alpha channels should never be negative, but clamping the value to 0..1 does not help.

The other commit only optimizes the font handling code. It's not required to fix #1491.

jlindgren90:
I'd like to understand why draw_text is called with a < 0 in the first place. That shouldn't happen.

radioactiveman:
It's because of using floating point math. I think comparing against 0 in `ui_infoarea_do_fade()` is not correct. Using an epsilon avoids -0.1 values.

The early return in `draw_text()` is still needed though to fix the emoji issue. But we can use an epsilon here as well.

```diff
diff --git a/src/gtkui/ui_infoarea.cc b/src/gtkui/ui_infoarea.cc
index c59f35a33..31a090031 100644
--- a/src/gtkui/ui_infoarea.cc
+++ b/src/gtkui/ui_infoarea.cc
@@ -17,6 +17,7 @@
  * the use of this software.
  */
 
+#include <float.h>
 #include <math.h>
 #include <string.h>
 
@@ -140,7 +141,7 @@ G_GNUC_END_IGNORE_DEPRECATIONS
 static void draw_text (GtkWidget * widget, cairo_t * cr, int x, int y, int
  width, float r, float g, float b, float a, int font_size, const char * text)
 {
-    if (a < 0)
+    if (a <= FLT_EPSILON)
         return;
 
     cairo_move_to (cr, x, y);
@@ -285,7 +286,7 @@ static void ui_infoarea_do_fade (void *)
         done = false;
     }
 
-    if (area->last_alpha > 0)
+    if (area->last_alpha > FLT_EPSILON)
     {
         area->last_alpha -= 0.1;
         done = false;

```

jlindgren90:
How about this? https://github.com/jlindgren90/audacious-plugins/commit/f754d970fc8b42bb67706107802eec3b5a48a1e5

jlindgren90:
Interesting, let's add a zero check as a workaround like you suggested then.

jlindgren90:
LGTM, thanks

PR #173, opened 2024-11-03
opus: add support for RG tags again

Almoped:
Re-added Thomas's original RG code and some logic to support both R128 and RG, while preferring R128.

### DIFF
--- a/src/opus/opus.cc
+++ b/src/opus/opus.cc
@@ -237,12 +237,13 @@ static bool update_replay_gain(OggOpusFile * opus_file,
             rg_info->album_peak = str_to_double(album_peak);
             rg_info->track_peak = str_to_double(track_peak);
         }
+        
+        AUDDBG("Album peak: %s (%f)\n", album_peak, rg_info->album_peak);
+        AUDDBG("Track peak: %s (%f)\n", track_peak, rg_info->track_peak);
     }
 
     AUDDBG("Album gain: %s (%f)\n", album_gain, rg_info->album_gain);
-    AUDDBG("Track gain: %s (%f)\n", track_gain, rg_info->track_gain);
-    AUDDBG("Album peak: %s (%f)\n", album_peak, rg_info->album_peak);
-    AUDDBG("Track peak: %s (%f)\n", track_peak, rg_info->track_peak);
+    AUDDBG("Track gain: %s (%f)\n", track_gain, rg_info->track_gain);    
 
     return true;
 }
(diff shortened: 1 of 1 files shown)


src/opus/opus.cc
```diff
@@ -181,6 +181,15 @@ static bool update_replay_gain(OggOpusFile * opus_file,
...
     const char * track_gain = opus_tags_query(tags, "R128_TRACK_GAIN", 0);
+    bool got_r128 = true;
+
+    /* try to read RG tags if R128 didn't work*/
+    if (!album_gain && !track_gain)
+    {
+        got_r128 = false;
+        album_gain = opus_tags_query(tags, "REPLAYGAIN_ALBUM_GAIN", 0);
+        track_gain = opus_tags_query(tags, "REPLAYGAIN_TRACK_GAIN", 0);        
+    }
```
radioactiveman, line 192: I would prefer this approach:

```c++
bool has_r128_tags = (album_gain && track_gain);

/* try to read RG tags if R128 didn't work */
if (!has_r128_tags)
{
    album_gain = opus_tags_query(tags, "REPLAYGAIN_ALBUM_GAIN", 0);
    track_gain = opus_tags_query(tags, "REPLAYGAIN_TRACK_GAIN", 0);        
}
```

src/opus/opus.cc
```diff
@@ -192,14 +201,43 @@ static bool update_replay_gain(OggOpusFile * opus_file,
...
+            if (!track_peak)
+                track_peak = album_peak;
+
+            rg_info->album_peak = str_to_double(album_peak);
+            rg_info->track_peak = str_to_double(track_peak);
+        }
+    }
 
     AUDDBG("Album gain: %s (%f)\n", album_gain, rg_info->album_gain);
     AUDDBG("Track gain: %s (%f)\n", track_gain, rg_info->track_gain);
```
radioactiveman, line 243: Bring back the peak log messages as well. They were removed with https://github.com/audacious-media-player/audacious-plugins/commit/78f33f7316de481d6ce34ed3614076fda700aa1c.

radioactiveman (changes requested):

src/opus/opus.cc
Almoped: That makes sense. I was unsure of a clear way to have it.

Almoped:
Trying to make friends with vscode and git it seems to be resolved now.

PR #174, opened 2024-11-22
Make UI appear on Qt::ApplicationActive for MacOS

jkl1337:
If the status icon plugin is used with close to system tray the UI main window is closed. On Mac OS there is not an easy way to make it reappear. The system tray activation trigger is disabled on Mac OS. This is for good reason due to one button mouse support and that this usually is intended to open the status/tray menu. Instead we can handle the application state change activate signal which will fire when the dock icon is clicked.

### DIFF
--- a/src/qtui/main_window.cc
+++ b/src/qtui/main_window.cc
@@ -21,6 +21,7 @@
 
 #include <libaudcore/drct.h>
 #include <libaudcore/i18n.h>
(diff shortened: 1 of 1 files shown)


radioactiveman:
Thanks, merged with minor adjustments.

I assume the Winamp interface (found in `skins-qt` folder) needs the same fix.
Could you please check and send another PR for it as well?

PR #176, opened 2024-11-29
pipewire: Revise plugin code. Closes: #1401

radioactiveman:
- Use RingBuf from audcore
- Respect output_buffer_size setting
- Improve get_delay() and drain() functions
- Add support for 5 and 7 channels

### DIFF
--- a/src/pipewire/pipewire.cc
+++ b/src/pipewire/pipewire.cc
@@ -21,6 +21,7 @@
  */
 
 #include <cmath>
+#include <ctime>
 
 #include <pipewire/pipewire.h>
 #include <spa/param/audio/format-utils.h>
@@ -28,10 +29,28 @@
 
 #include <libaudcore/i18n.h>
 #include <libaudcore/plugin.h>
+#include <libaudcore/ringbuf.h>
 #include <libaudcore/runtime.h>
 
 #if !PW_CHECK_VERSION(0, 3, 33)
-  #define PW_KEY_NODE_RATE "node.rate"
+#define PW_KEY_NODE_RATE "node.rate"
+#endif
+
+#if !PW_CHECK_VERSION(0, 3, 50)
+static inline int pw_stream_get_time_n(struct pw_stream * stream,
+                                       struct pw_time * time, size_t size)
+{
+    return pw_stream_get_time(stream, time);
+}
+#endif
+
+#if !PW_CHECK_VERSION(1, 0, 4)
+static uint64_t pw_stream_get_nsec(struct pw_stream * stream)
+{
+    struct timespec ts;
+    clock_gettime(CLOCK_MONOTONIC, &ts);
+    return SPA_TIMESPEC_TO_NSEC(&ts);
+}
 #endif
 
 class PipeWireOutput : public OutputPlugin
@@ -103,9 +122,8 @@ private:
     int m_aud_format = 0;
     int m_core_init_seq = 0;
 
-    unsigned char * m_buffer = nullptr;
-    unsigned int m_buffer_at = 0;
-    unsigned int m_buffer_size = 0;
+    RingBuf<unsigned char> m_buffer;
+    unsigned int m_pw_buffer_size = 0;
     unsigned int m_frames = 0;
     unsigned int m_stride = 0;
     unsigned int m_rate = 0;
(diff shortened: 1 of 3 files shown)


radioactiveman:
@jlindgren90: Is this looking good for you? Especially the implementation of `get_delay()`? SDL and qtaudio for example store a paused flag and use it there. Should we do the same for PipeWire?

maris-ab:
Now checking pw_stream_get_time_n() function, maybe time diff is useful for delay calculation, will came soon with decision.

src/pipewire/pipewire.cc
```diff
@@ -103,8 +104,9 @@ class PipeWireOutput : public OutputPlugin
     int m_aud_format = 0;
     int m_core_init_seq = 0;
 
-    unsigned char * m_buffer = nullptr;
-    unsigned int m_buffer_at = 0;
+    timespec m_time {};
```
jlindgren90, line 107: `m_time` is not very descriptive. Can we use a better name, or at least add a comment indicating what this value means?

src/pipewire/pipewire.cc
```diff
@@ -159,36 +161,61 @@ void PipeWireOutput::pause(bool pause)
...
+
+    // Get time difference from last buffer fill
+    timespec time_now {};
+    clock_gettime(CLOCK_REALTIME, &time_now);
+
+    int time_diff =
+        (((time_now.tv_sec - m_time.tv_sec) * 1000000000) +
+         (time_now.tv_nsec - m_time.tv_nsec)) / 1000000;
+
+    if (time_diff < 0)
```
jlindgren90, line 180: `aud::clamp` would be convenient for the range-limiting.

src/pipewire/pipewire.cc
```diff
@@ -159,36 +161,61 @@ void PipeWireOutput::pause(bool pause)
...
 {
     pw_thread_loop_lock(m_loop);
-    if (m_buffer_at > 0)
-        pw_thread_loop_timed_wait(m_loop, 2);
+
+    int buflen;
+    while ((buflen = m_buffer.len()) > 0)
+    {
+        pw_thread_loop_timed_wait(m_loop, 1);
+        if (buflen <= m_buffer.len())
```
jlindgren90, line 196: This part looks a little suspect. Are we waiting 1ms, and then expecting the buffer level to have decreased? That seems like it will break if PipeWire uses a period greater than 1ms.

Is there no PipeWire API to a period-based (rather than time-based) wait?

src/pipewire/pipewire.cc
```diff
@@ -159,36 +161,61 @@ void PipeWireOutput::pause(bool pause)
...
+        {
+            AUDERR("PipeWireOutput: buffer drain lock\n");
+            break;
+        }
+    }
 
     pw_stream_flush(m_stream, true);
-    pw_thread_loop_timed_wait(m_loop, 2);
+    pw_thread_loop_timed_wait(m_loop, 1);
```
jlindgren90, line 204: What is this extra `timed_wait` for? Seems like we could use an explanatory comment here.

src/pipewire/pipewire.cc
maris-ab: From PW doc it is waiting 1 second. Checked also PW source code thread-loop.c:
int pw_thread_loop_timed_wait(struct pw_thread_loop *loop, int wait_max_sec) ...
w_thread_loop_get_time(loop,&timeout, wait_max_sec * SPA_NSEC_PER_SEC)

src/pipewire/pipewire.cc
maris-ab: Thanks for advice, corrected in next version.

src/pipewire/pipewire.cc
maris-ab: Removed m_time in next version as using pw_stream_get_time_n for getting delay.

src/pipewire/pipewire.cc
jlindgren90: Okay that's fine then. Thanks for checking.

src/pipewire/pipewire.cc
maris-ab: pw_thread_loop_timed_wait(m_loop, 1) is required after pw_stream_flush(m_stream, true);
otherwise on_drained(void * data) is not triggered.
Will be comment in code on that.

jlindgren90:
Looking very nice. @radioactiveman any idea why the CI build fails on Ubuntu 22.04?

src/pipewire/pipewire.cc
```diff
@@ -164,31 +178,60 @@ void PipeWireOutput::pause(bool pause)
...
+
+#if PW_CHECK_VERSION(0, 3, 50)
+        add_delay += time.buffered * 1000 / m_rate;
+#endif
+        if (time.rate.denom > 0)
+            add_delay += time.delay * 1000 * time.rate.num / time.rate.denom;
+    }
+
+    return buff_time + pw_buff_time - time_diff + add_delay;
 }
```
radioactiveman, line 202: @maris-ab: `pw_time.buffered` is only available since 0.3.50.
I have split `add_delay = (v_time.buffered + v_time.queued) * 1000 / m_rate;` into two separate calculations.

Also changed
`time_diff = aud::clamp<int>(0, time_diff, pw_buff_time);`
to
`time_diff = aud::clamp(time_diff, 0, pw_buff_time);`

radioactiveman:
> Looking very nice. @radioactiveman any idea why the CI build fails on Ubuntu 22.04?

The pipeline is green now. I needed to add version checks and fallback methods for Ubuntu 22.04.
Thanks again @maris-ab for your patches. :+1:

jlindgren90 (approved):

PR #177, opened 2025-02-01
pipewire: fall back to other output plugins if not available

jlindgren90:
This makes Audacious again work "out of the box" on systems without pipewire installed/running.

@radioactiveman Can you check that this doesn't break systems that *do* have pipewire available?

### DIFF
--- a/src/pipewire/pipewire.cc
+++ b/src/pipewire/pipewire.cc
@@ -312,6 +312,17 @@ bool PipeWireOutput::init()
 {
     aud_config_set_defaults("pipewire", defaults);
     pw_init(nullptr, nullptr);
+
+    /* try to create main loop to test if pipewire is available */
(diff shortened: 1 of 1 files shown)


mschwendt:
Works here with Fedora 41 Audacious 4.4.2. Tested with (1) current configuration set to Pipewire, and (2) switching between Pipewire and PulseAudio while playing.

radioactiveman:
I can confirm the same, thanks. Should we revert commit https://github.com/audacious-media-player/audacious-plugins/commit/434c518c513b60738dea92dfee9db80c5dd8b6f1 now?

jlindgren90:
> I can confirm the same, thanks. Should we revert commit [434c518](https://github.com/audacious-media-player/audacious-plugins/commit/434c518c513b60738dea92dfee9db80c5dd8b6f1) now?

Sure, fine with me

PR #179, opened 2025-02-19
Fix track length for CUE files specifying "INDEX 00" pregap

mschwendt:
A followup to issue #1538. Libcue takes over calculating the correct track length under consideration of any pregap.

### DIFF
--- a/src/cue/cue.cc
+++ b/src/cue/cue.cc
@@ -161,9 +161,9 @@ bool CueLoader::load (const char * cue_filename, VFSFile & file, String & title,
 
             if (same_file)
             {
-                int len = (int64_t) track_get_length (cur) * 1000 / 75;
-                tuple.set_int (Tuple::EndTime, begin + len);
-                tuple.set_int (Tuple::Length, len);
+                int length = (int64_t) track_get_length (cur) * 1000 / 75;
+                tuple.set_int (Tuple::EndTime, begin + length);
+                tuple.set_int (Tuple::Length, length);
             }
             else
             {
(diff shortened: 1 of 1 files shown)


radioactiveman:
Thanks for working on this Michael. :+1:

The last related code change was commit https://github.com/audacious-media-player/audacious-plugins/commit/9fa6215267f63b756840cde9a0a9d5026906f6e9. Could you please verify that there is no regression for issue [651](https://redmine.audacious-media-player.org/issues/651)?

radioactiveman:
@jlindgren90: What is your recommendation? I am literally c(l)ueless here. :wink:

My choice would be to approach the issue which is more common if there is no 100 % fix. Or add a settings option to the plugin to support both cases.

jlindgren90:
From what I remember, the "pregap" between tracks is the portion where an actual CD player would display a negative time counting toward 0:00 at the start of the next track.

We don't have logic to support that properly (everything in the code assumes tracks start at 0:00) so we have to get something wrong.

From worst to best option (IMHO):

1. skip the pre-gap audio altogether
2. make the pre-gap part of the following track, and shift the displayed time counter (so we display 0:00 where it should be -0:03 or whatever)
3. make the pre-gap audio part of the preceding track

Either option 2 or 3 would result in longer displayed track lengths than the "correct" length.

I think option 3 is the current behavior (or at least that was the intent) and I would vote to keep that behavior.

mschwendt:
Option 3 is current behaviour and a plausible, acceptable default. Play from start of current track to start of next track (including its pregap).

Option 2 would be wrong, because it would not start a track at INDEX 01 like a CD player does when selecting a specific track. For the case of starting playback of an individual track (skipping to it), option 2 would start at INDEX 00, and if it's not only 1-2 seconds of silence but something audible at the end a previous track, it's likely to sound worse than option 3.

PR #180, opened 2025-03-28
Add MIME type audio/prs.sid

mschwendt:
Support the official MIME type for SID module files. It has replaced the inofficial audio/x-psid and audio/x-sidtune types.

### DIFF
--- a/src/sid/xmms-sid.cc
+++ b/src/sid/xmms-sid.cc
@@ -38,6 +38,7 @@ class SIDPlugin : public InputPlugin
 {
 public:
     static const char *const exts[];
+    static const char *const mimes[];
 
     static constexpr PluginInfo info = {
         N_("SID Player"),
@@ -48,6 +49,7 @@ public:
 
     constexpr SIDPlugin() : InputPlugin(info, InputInfo(FlagSubtunes)
         .with_priority(5) /* medium priority (slow to initialize) */
+        .with_mimes (mimes)
         .with_exts(exts)) {}
 
     bool init()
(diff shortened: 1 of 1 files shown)


radioactiveman:
If I am not missing something, only the Neon plugin actually implements `VFSImpl::get_metadata("content-type")`. Reading the MIME type of a local file will always return an empty string. And since a SID file is usually not streamed via network I assume your change has practically no effect at all.

Also `file --mime-type sample.sid` results in `application/octet-stream` here for SID files I found online.

@jlindgren90: Did I understand this correctly? Is the change valid anyway?

mschwendt:
Well, there have been (and perhaps still are) web servers handing out sidtunes as "Content-Type: audio/prs.sid" (and earlier, the older audio/x-sidtune and audio/x-psid types).  It's a niche market plugin, sure. No worries.

But I cannot foresee how useful a more complete set of MIME types might become eventually.

Btw, here's the registered entry: https://www.iana.org/assignments/media-types/audio/prs.sid
(from: https://www.iana.org/assignments/media-types/media-types.xhtml )

mschwendt:
@radioactiveman A potential use-case would be to https://github.com/audacious-media-player/audacious/issues/1583
The probing step knows the stream's content-type, but if no plugin declares that that type, the prober still tries any input plugin in search of one that accepts the stream. Knowing which plugins **don't** support the content-type (because they advertize other types only), the prober could skip those plugins. Makes sense?

jlindgren90:
I think it's fine to add the mime-type even if it's a niche use-case.

radioactiveman:
OK, merged now. Thanks Michael for the PR.

PR #182, opened 2025-07-01
gtkui/qtui: Include extension in file name playlist column

radioactiveman:
See also: https://github.com/orgs/audacious-media-player/discussions/1610


jlindgren90:
LGTM, thanks

PR #183, opened 2025-07-27
gl-spectrum: Add modern GL renderer and GtkGLArea support

### LINKED ISSUE #1546: Use GtkGLArea for OpenGL Spectrum Analyzer plugin
**Description**

Porting the `glspectrum` plugin to use `GtkGLArea` (available since GTK 3.16) would be beneficial for two reasons:

- No need of different code for X11 and Windows as GTK would take care of it internally
- Wayland support, the current code for Linux requires X11/XWayland explicitly

As my OpenGL knowledge is not good enough to port this plugin myself, I file this here as enhancement. Patches are welcome. :slightly_smiling_face:

**Additional information**
1. https://docs.gtk.org/gtk3/class.GLArea.html
2. https://www.bassi.io/articles/2015/02/17/using-opengl-with-gtk/
3. https://github.com/ebassi/glarea-example

CC: @CBke, @carlo-bramini as you have been contributors for this plugin

rsxrwscjpzdzwpxaujrr:
Closes https://github.com/audacious-media-player/audacious/issues/1546

[Video demonstration](https://www.youtube.com/watch?v=XhhnwEC9pWI)

AFAIU gl-spectrum plugin using GLX is one of the blockers for transitioning to using wayland by default. Well, not anymore, because I implemented GL context creation through GtkGLArea widget. Unfortunately, there are no good ways to force GtkGLArea to create a legacy GL context, it only falls back to legacy context if creating a core GL context failed, and this plugin used only legacy GL functions for rendering. So I implemented a new renderer which uses modern GL functions. The modern renderer is not uploading the whole scene to GPU on every frame like the legacy one does, but uploads it only once and then uploads only the height map on every frame, which is used calculate the actual geometry in a vertex shader, which should make it faster. The context version is being checked in runtime and the modern renderer is used if requirements are met, if not it's just using the old one. The modern renderer is also used by default in the case of context creation through GLX and Win32. If everything works fine, GLX and Win32 GL context creation code can be safely removed to reduce unnecessary complexity, and it won't change the minimal required OpenGL version as we can always fall back to the legacy rendering.

I also want to note that with the modern renderer cool effects can be added in the future. For now it's just replicating the old one.

Please test in on Windows and macOS before merging, as I can't verify if it's working there (it should).

### DIFF
--- a/.github/actions/install-dependencies/action.yml
+++ b/.github/actions/install-dependencies/action.yml
@@ -29,12 +29,14 @@ runs:
           mingw-w64-ucrt-x86_64-flac
           mingw-w64-ucrt-x86_64-fluidsynth
           mingw-w64-ucrt-x86_64-gcc
+          mingw-w64-ucrt-x86_64-glm
           mingw-w64-ucrt-x86_64-gtk2
           mingw-w64-ucrt-x86_64-json-glib
           mingw-w64-ucrt-x86_64-lame
           mingw-w64-ucrt-x86_64-libbs2b
           mingw-w64-ucrt-x86_64-libcdio-paranoia
           mingw-w64-ucrt-x86_64-libcue
+          mingw-w64-ucrt-x86_64-libepoxy
           mingw-w64-ucrt-x86_64-libmodplug
           mingw-w64-ucrt-x86_64-libopenmpt
           mingw-w64-ucrt-x86_64-libsamplerate
--- a/.github/actions/install-dependencies/install-dependencies.sh
+++ b/.github/actions/install-dependencies/install-dependencies.sh
@@ -24,14 +24,16 @@ ubuntu_packages='gettext libadplug-dev libasound2-dev libavformat-dev
                  libmpg123-dev libneon27-gnutls-dev libnotify-dev libopenmpt-dev
                  libopusfile-dev libpipewire-0.3-dev libpulse-dev
                  libsamplerate0-dev libsdl2-dev libsidplayfp-dev libsndfile1-dev
-                 libsndio-dev libsoxr-dev libvorbis-dev libwavpack-dev libxml2-dev'
+                 libsndio-dev libsoxr-dev libvorbis-dev libwavpack-dev libxml2-dev
+                 libepoxy-dev libglm-dev'
 
 ubuntu_qt5_packages='libqt5opengl5-dev libqt5svg5-dev libqt5x11extras5-dev
                      qtbase5-dev qtmultimedia5-dev'
 ubuntu_qt6_packages='qt6-base-dev qt6-multimedia-dev qt6-svg-dev'
 
 macos_packages='adplug faad2 ffmpeg libbs2b libcue libmms libmodplug libnotify
-                libopenmpt libsamplerate libsoxr neon opusfile sdl3 wavpack'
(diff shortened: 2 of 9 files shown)


radioactiveman:
Thanks for working on this. :+1: I'm looking forward to try it out when finding some free time.

> AFAIU gl-spectrum plugin using GLX is one of the blockers for transitioning to using wayland by default.

The main reason for using XWayland by default is that the Winamp interface is unusable on Wayland, see [#1560](https://github.com/orgs/audacious-media-player/discussions/1560) for more details. Supporting the gl-spectrum plugin on Wayland is of course still beneficial.

Questions from a first quick look:
- We already require GTK 3.18 if GTK3 is enabled. And these versions depend on libepoxy. Maybe we can simplify the code and build system checks if we say "always use the modern renderer for GTK3"?
- glm however would be a new dependency. Is it possible to avoid this library?

radioactiveman:
Audacious 4.6 no longer enables Xwayland by default for the GTK/Qt interface. Therefore making the plugin compatible with Wayland became more important now.

Have you found a solution to fix the performance issue? Are you interested in finishing your PR?

rsxrwscjpzdzwpxaujrr:
The PR is pretty much finished, I can resolve the conflicts and it'll be ready to be merged if you're. The performance issue is due to how GtkGLArea in GTK3 is implemented and I don't think it's something that we can fix. We can 
1. Always use GtkGLArea on GTK3 and keep GLX context creation only for GTK2 to keep the code clean (the way it's currently implemented in the PR, but no code is removed and you still can get old context creation with `-DUSE_GTKGLAREA=OFF`)
2. Use GLX on X/Xwayland and WGL on Windows and keep GtkGLArea on Wayland. The code can be changed do to determining in runtime whether to use GtkGLArea or GLX/WGL instead of compilation flags like it's currently in the PR.

Which way do you want me to proceed? The performance regression is negligible and probably will be noticed only on pre-2010 tech.

Also, there are two renderers now, as we can't ask GTK3 to give us legacy OpenGL I created a renderer working with a more modern OpenGL. Which way do you want me to proceed? 
1. Keep both renderers, the plugin will fall back to the legacy one when it can't run the modern one. (This is how it's currently implemented in the PR)
2. Get rid of the legacy renderer entirely to keep the code clean and use the modern renderer in both GTK3 and GTK2, the downside is that some pre-2008 tech won't be able to run it anymore.

Also, are you sure that you want me to proceed with getting rid of the glm dependency? It's a header-only dependency. If you are, I will get rid of it.

I'm not comfortable with making these decisions myself as this is not my project so I'm waiting for your response to proceed.

radioactiveman:
My suggestion would be:
* Always use GtkGLArea with GTK 3, no fallback -> get rid of USE_GTKGLAREA completely
* Keep the old renderer for GTK 2 only, do not use the modern renderer for GTK 2

Then it probably would make sense to just split the plugin into two implementations. One for GTK 3, and one for GTK 2. Which one gets compiled can be decided in meson.build.

This makes the code much more readable. It already has too many ifdef's and with your proposed changes this gets even worse.

Could you please prepare a minimal GTK 3 only implementation to see if this is feasible?

Regarding glm: How much custom code would be required when we want to avoid glm? If it is let's say ~100 lines, I would be in favor of that. But I can't estimate that yet.

Thanks.

PR #185, opened 2025-08-02
opus: decode METADATA_BLOCK_PICTURE manually

bkoropoff:
I was trying to track down why Audacious was choking on thumbnails in some of my opus files.  It turns out the `libopusinfo` implementation of tag parsing is [strict about some of the fields](https://github.com/xiph/opusfile/blame/main/src/info.c#L655) in the `METADATA_BLOCK_PICTURE` and will return an error even if it contains a usable image.  Some common programs seem to write tags with this problem, and most applications that read thumbnails from Ogg containers (including the Audacious Vorbis plugin) seem to be lenient about it.  I therefore just copied the code over from the Vorbis plugin.

### DIFF
--- a/src/opus/meson.build
+++ b/src/opus/meson.build
@@ -5,7 +5,7 @@ have_opus = opusfile_dep.found()
 if have_opus
   shared_module('opus',
     'opus.cc',
-    dependencies: [audacious_dep, opusfile_dep],
+    dependencies: [audacious_dep, opusfile_dep, glib_dep],
     name_prefix: '',
     include_directories: [src_inc],
     install: true,
--- a/src/opus/opus.cc
+++ b/src/opus/opus.cc
@@ -22,11 +22,15 @@
 
 #include <opus/opusfile.h>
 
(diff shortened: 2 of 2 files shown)


bkoropoff:
Whoops, should have known that was a smart pointer and not just a weird typedef.  Fixed the memory leak.

I know yt-dlp also attaches cover art with bad metadata.  Mutagen (which yt-dlp uses) should probably enforce the invariant that libopusinfo expects, but it doesn't have a bespoke interface for embedded pictures for generic Ogg containers, only FLAC files.  Every other music player I've looked at (Rhythmbox, VLC, etc.) avoids using libopusinfo for this task presumably because of bad tags "in the wild", not to mention the `opusinfo` program itself.

radioactiveman:
I have opted now to keep the opusfile image parsing, but fallback to manual decoding if an image tag was rejected.

See https://github.com/audacious-media-player/audacious-plugins/commit/07056ddde5705ee34d1c90fdb2609883a453c9a6, thanks for your groundwork and the PR. :+1:

PR #186, opened 2025-10-01
skins & skins-qt: visualizer rework

0x5066:
This PR aims to rework the visualizer seen in the Winamp Classic Interface.

___

## Motivation

Reimplementations of the Winamp Classic UI, like Webamp, have studied and reimplemented all sorts of small details that makes classic skins look and feel as though you were actually using Winamp, the downside is that Webamp only exists in the browser and is not a fully fledged media player.

Though people still have a desire to re-experience those skins again, and more often than not, they install Audacious, which does a pretty decent job of emulating classic skins, but it could be better in some areas. This PR touches on the visualizer aspect of it.

<details>

<summary>Audacious vs. my patch series comparison</summary>

comparison

The top blocks shows Audacious' skins-qt (on the left) in its current state, next to skins-qt with my changes.
The windowshaded shows the lines mode in the same fashion, skins-qt with my patches applied at the top, current git below it, the same goes for the solid mode.

Spectrum analyzer comparison, the top window is more stable and smooth, whereas the bottom is more jittery and unstable:
ezgif-3836ffc9fd4427

In the windowshaded mode, the analyzer used to not have any sort of falloff applied and was instead very reactive, it was also missing peaks:
ezgif-3b380af9f69f0b

The same applies to the VU Meter (just that it isn't missing the peaks, they weren't there to begin with in Winamp):
ezgif-3c2565f45b40db

</details>

## Codebase changes

The changes were only done in ``vis.cc``, ``svis.cc`` and ``vis.h`` (in both ``skins-qt`` and ``skins`` respectively), the focus was on improving the rendering styles for the Oscilloscope and Analyzer modes, with additions and refactoring done where necessary.
This also included a bugfix, where for the longest time, the oscilloscope had both the wrong color order, as well as having one of the colors from the analyzer incorrectly assigned to it.

The line mode for the scope was improved based on observations of Winamp and Webamp, as well as the solid mode, the dots mode remained unchanged.
This improvement was also applied to the mini visualizer, with the weird line drawing that's done and it only using one color for the overall scope, the scale for the audio data was also adjusted, as it visually appeared "too loud".

___

This is my second attempt at wanting to improve Audacious' Winamp skins support which is more focused and (hopefully) more organized, I will admit though that I did tend to forget to update ``skins`` as I primarily was using and testing with the Qt version, but from my own testing, they should at least line up with what I was seeing on my end.

I hope that this PR doesn't cause too much trouble.

### DIFF
--- a/src/skins-qt/svis.cc
+++ b/src/skins-qt/svis.cc
@@ -32,7 +32,6 @@
 #include "vis.h"
 
 static const int svis_analyzer_colors[] = {14, 11, 8, 5, 2};
-static const int svis_scope_colors[] = {20, 19, 18, 19, 20};
 static const int svis_vu_normal_colors[] = {16, 14, 12, 10, 8, 6, 4, 2};
 
 #define RGB_SEEK(x,y) (set = rgb + 38 * (y) + (x))
--- a/src/skins-qt/svis.cc
+++ b/src/skins-qt/svis.cc
@@ -26,6 +26,7 @@
 
 #include <string.h>
 #include <libaudcore/objects.h>
+#include <libaudcore/drct.h>
 
 #include "skins_cfg.h"
 #include "skin.h"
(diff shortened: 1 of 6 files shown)


0x5066:
Sure! I can do that.
There's two bugs here, though realistically one, one bug is the incorrect color/color order for the oscilloscope, the other being that the scope isn't perfectly centered (see attached)
image
which might be something in libaudcore as I did notice the polarity of the audio being wrong in both the classic skin interface and the included Blur Scope plugin, and thus out of scope for this PR

I can address the color thing no problem, I'm not sure about the formatting issues as I did try to keep in line with the pre-existing formatting and code style.

PR #181, opened 2025-06-13
Add toggle buttons for 'No Playlist Advance'

sunhaitao:
This pull request adds the toggle buttons in both the Qt and GTK UIs for 'No Playlist Advance'. It makes the current play policy more clear.

This depends on <https://github.com/audacious-media-player/audacious/pull/1618>.

PR #187, opened 2025-10-03
Add toggle buttons for 'No Playlist Advance' using ① (U+2460)

jlindgren90:
Alternative approach to:
- #181

Unicode ① (U+2460) seems like a suitable graphic for the buttons and avoids having to add new icons, which would not fit in with the system icon theme anyway.

I also changed the button order (repeat, shuffle, no playlist advance) and fixed up the gtkui implementation, which wasn't updating the toggle button properly at startup or when toggled via menu/hotkey.


jlindgren90:
@radioactiveman care to review this?

It makes the UI to play/repeat a single track much more visible and is a nice improvement IMHO.

Ideally there would be a standard themed icon for this, but I don't know of one.

radioactiveman:
> Maybe we just forget it for now? The option is already there in the menu, and power users who care enough about the feature can use the keyboard shortcut.

OK, sounds reasonable to me.

PR #188, opened 2025-10-08
filesystem-qt: Introduce new plugin to manage local audio files

radioactiveman:
It basically serves as an embedded file manager view
- allowing to create or extend playlists
- opening the music folders and cover art
- supporting drag and drop

Depends on the new libaudcore function added in https://github.com/audacious-media-player/audacious/pull/1671.

Supersedes PR https://github.com/audacious-media-player/audacious-plugins/pull/184 by @hdijkema with a simpler and more straightforward approach.


jlindgren90:
Some issues I've noticed in a quick test:
- The quick search only finds items in the base folder and doesn't seem to be recursive at all (even into folders that are visually expanded). So it's not actually that useful.
- Sometimes when expanding a folder, multiple rows get expanded. It seems random; I can't reproduce it consistently. Could be a Qt bug.

hdijkema:
When I try this on my linux Mint distribution (Linux Mint 22.1 Xia, Cinnamon) and when I try to select a different music library, it gets a malloc error:

- malloc_consolidate(): unaligned fastbin chunk detected

and exits.

I've tried to add some Debug() << statements, but they don´t show up. Don't know where to look for logging.

radioactiveman:
If you use `aud::log()` you need to call Audacious with "-VV" parameter for verbose debug logging. Or simply use cout/printf if you don't want to use gdb.

> when I try to select a different music library, it gets a malloc error.

So the initial default path (your Music / Home directory) works but then selecting a different path with `Change Music Folder` Audacious crashes? Is there anything special with this directory? `setCurrentDirectory()` is probably the first method to debug I guess.

hdijkema:
It does work when I manually set my own music folder in the audacious config file.

hdijkema:
Even if I choose the same folder, I get this:

```
hans@xenamedes:~/src/audacious-plugins/src/filebrowser-qt$ audacious
WARNING filebrowser-qt.cc:298 [changeMusicDirectory]: changeMusicDir
WARNING filebrowser-qt.cc:301 [changeMusicDirectory]: changeMusicDir: /muziek
WARNING filebrowser-qt.cc:305 [changeMusicDirectory]: folder Choosen: /muziek
WARNING filebrowser-qt.cc:307 [changeMusicDirectory]: setCurrentDirectory executed: /muziek
malloc(): unaligned tcache chunk detected
Afgebroken (geheugendump gemaakt)
hans@xenamedes:~/src/audacious-plugins/src/filebrowser-qt$ 
```

radioactiveman:
On Ubuntu 24.04 with Qt 6.4.2 I could reproduce the crash when using "Change Music Folder". Seems to be related to the deletion of the context menu. Without `menu->setAttribute(Qt::WA_DeleteOnClose);` in `contextMenuEvent()` there is no crash.

PR https://github.com/audacious-media-player/audacious-plugins/pull/190 removed the "Change Music Folder" entry.

@jlindgren90: Can you please double-check this? There is no crash anymore without the dialog. But since `contextMenuEvent()` still uses `Qt::WA_DeleteOnClose` (similar to search-tool-qt) I'm not sure if it is correct. Thanks.

jlindgren90:
Not at my computer right now to verify, but: QFileDialog::getExistingDirectory is a blocking call (runs a nested event loop, I think). Probably after that returns, something still referenced the QAction or the parent context menu, which had both been deleted at that point.

Nested event loops are tricky that way - I've tried to avoid using them in Audacious.

PR #201, opened 2026-05-04
skins-qt: Accept dropped files on main window

u-foka:
The files are handled similar as opening from the menu respecting the "Open files in temporary playlist" setting.

This change is more or less copying the original winamp behaviour (while respecting audacious playlist handling)
Partially addressing https://github.com/audacious-media-player/audacious/issues/1064

(First time contributor here, please just let me know if I overlooked something)

### DIFF
--- a/src/skins-qt/main.cc
+++ b/src/skins-qt/main.cc
@@ -60,6 +60,9 @@
 #include "vis.h"
 #include "view.h"
 
+#include <QDragEnterEvent>
+#include <QMimeData>
+
 #define SEEK_THRESHOLD 200 /* milliseconds */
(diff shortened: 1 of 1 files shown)


radioactiveman:
Looks good to me, thanks for your first contribution. :+1:

From my understanding the feature request #1064 is unrelated to the Winamp interface. So it is correct to not mention it in the commit message.

If you would like to improve the Qt Winamp interface even more, you can add handling of skin files via Drag and Drop.
The GTK version automatically imports and selects the skin (.wsz or .zip) when dropping the file on the main window. Feel free to submit another pull request for this.

radioactiveman (approved):

PR #203, opened 2026-06-16
allow dropping skin files on main window of skins interface

u-foka:
As suggested in my previos PR, this one allows dropping skin files onto the main window of the skins interface.
I wasn't sure what the original winamp behaviour was, so I went with the following:
- Skins have to appear on the toplevel list of dropped files as .zip or .wsz
- Multiple skins can be imported at once, and the last one will be activated
- Skins can be mixed with media files, in which case the skins are pulled from the list of dropped files list, and the rest of the files will be loaded assuming they're media as previously

There is one known issue: If the settings window is open while the skins are being dropped, the skin list won't be updated.

### DIFF
--- a/src/skins-qt/main.cc
+++ b/src/skins-qt/main.cc
@@ -55,12 +55,11 @@
 #include "number.h"
 #include "playlist-widget.h"
 #include "playstatus.h"
+#include "skin.h"
 #include "textbox.h"
 #include "window.h"
 #include "vis.h"
 #include "view.h"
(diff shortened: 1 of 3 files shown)


radioactiveman:
Thank you for working on this and sorry for the late response, I have missed the notification for the PR.

Pushed now with slight modifications and further cleanups.

u-foka:
Thanks for cleaning up my mess ;)

PR #204, opened 2026-06-24
waveform visualization plugin

0000xFFFF:
waveform

- Adds a waveform visualization plugin for the currently playing track.
- Qt only.
- Uses libav to extract and generate waveform peak data.
- Caches generated waveform data in AudPath::UserDir/waveform_cache to avoid repeated processing.
- Supports mouse-based seeking directly from the waveform display.

#### Notes:

- The UI is still fairly rough and open to iteration; suggestions and changes are expected.
- Testing has been fairly limited so far (only Arch Linux); additional testing and feedback would be appreciated.
- May add zoom functionality in a future update.

https://www.redmine.audacious-media-player.org/issues/533
https://www.redmine.audacious-media-player.org/issues/936


0000xFFFF:
Added coloring for each band, low -> red, mid -> green, high -> blue. Quiet bands are faintly visible.

Here's a before and after:
image

PR #205, opened 2026-07-29
skins-qt: Prevent windows become offscreen when desktop geometry changes

u-foka:
Hi, so I had this issue when I removed the external monitor from my mac, the skins-qt windows went off the screen and I had zero chance of recovering them. this pr aims to fix that, both on startup (when the saved positions were off screen) and during runtime, when the external monitor is removed while audacious is running. There are some glitching when moving windows docked together from one display to another, that I couldn't figure out the reason of. Also it was only tested on macos, still trying to get a working linux build environment :)


radioactiveman:
Your approach is far too complex in my opinion. I do not think an application should have to monitor whether screens are added/removed and respond to that. Restoring the window position when starting the application should however take into account if the position is actually visible.

Please try this simple alternative:

Instead of calling `QWidget::move (int x, int y)` in the Window constructor use `QWidget::move (restorePosition (x, y, size ()))`.

```c++
QPoint restorePosition (int x, int y, const QSize & size)
{
    QRect windowRect (QPoint (x, y), size);

    for (QScreen * screen : QGuiApplication::screens ())
    {
        if (screen->availableGeometry ().intersects (windowRect))
            return QPoint (x, y);
    }

    return QGuiApplication::primaryScreen ()->availableGeometry ().center ();
}
```

@jlindgren90: What is your view on that? How would you fix this?

jlindgren90:
The check for "intersects available geometry" looks good. Rather than using a centered position as fallback, how about using the same default positions we would use on first install / new user account? That should keep the 3 windows (main, playlist, EQ) from all overlapping at screen center.

radioactiveman:
Valid point, users may still need to reposition the windows then. Our default playlist position (hidden though) is to the right of the main window. But users may also have docked the playlist below the main or equalizer window.

Maybe it's possible to calculate an x and y offset and then move the windows relatively (e.g. by 1000px to the left) to become visible on the primary screen, but keep their relation to each other (main, equalizer, playlist).

u-foka:
Hi, thanks for the feedback!

Gonna look into the default, and will check if runtime monitoring is necessary. You're right that the OS is supposed to move the windows automatically when the screen geometry changes during runtime.

Now I can also test on linux. I had to set up an environment as I'm working now on a new plugin for opensubsonic server connection :)

PR #206, opened 2026-07-29
console: check for companion m3u before opening

steckums:
The companion .m3u/.m3u8 lookup added in #198 attempts to open both paths unconditionally, so every file without a sidecar produces two VFS error messages in the log.

Test with VFSFile::test_file() first. Behavior is unchanged when a sidecar exists; also avoids opening the .m3u8 twice on the success path, since m3u_path was not reassigned in that case.


radioactiveman:
Looks good to me, thanks. :+1:

@ramonsmits: FYI

PR #208, opened 2026-08-04
sid: Implement seeking

mvtiaine:
there is no seek API in libsidplayfp, so use "brute force" seek

### DIFF
--- a/src/sid/xmms-sid.cc
+++ b/src/sid/xmms-sid.cc
@@ -66,9 +66,12 @@ public:
 
 private:
     bool delayed_init();
+    void seek(int seek_value);
 
+    Index<char> m_buffer;
     bool m_initialized = false;
     bool m_init_failed = false;
+    int64_t m_bytes_played = 0;
 };
 
 EXPORT SIDPlugin aud_plugin_instance;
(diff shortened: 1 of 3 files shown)


mvtiaine:
ubuntu fails because reset() is only available since 2.15:
https://github.com/libsidplayfp/libsidplayfp/blob/libsidplayfp-2.15/src/sidplayfp/sidplayfp.h#L143C9-L143C19

I'll add version guard if/when https://github.com/audacious-media-player/audacious-plugins/pull/207 is merged

radioactiveman:
How can I try this out? All SID sample files I found have no length and seeking is therefore not possible.

mvtiaine:
The songlengths support is kind of hidden feature :P
https://github.com/audacious-media-player/audacious-plugins/blob/master/src/sid/xs_sidplay2.cc#L177

It expects Songlengths.md5 file from HVSC (https://www.hvsc.c64.org) in `SIDDATADIR/sidplayfp` (along kernal/basic/chargen files from libsidplayfp), which resolves to `$prefix/share/sidplayfp` usually, so for example /usr/local/share/sidplayfp/Songlengths.md5.

I have it symlinked to the HVSC dir:

```
$ ls -la /usr/local/share/sidplayfp/Songlengths.md5 
lrwxr-xr-x  1 root  wheel  49 Jan 17  2023 /usr/local/share/sidplayfp/Songlengths.md5 -> /Users/tundrah/C64Music/DOCUMENTS/Songlengths.md5
```

src/sid/xs_sidplay2.h
```diff
@@ -5,12 +5,21 @@
 
 #include <stdint.h>
```
radioactiveman, line 6: Include `<sidplayfp/sidplayfp.h>` here for `LIBSIDPLAYFP_VERSION_MAJ` and `LIBSIDPLAYFP_VERSION_MIN`.

Without it, my IDE (zed) shows the wrong #if-else branch as active in the .cc files.

src/sid/xs_sidplay2.cc
```diff
@@ -207,6 +201,16 @@ void xs_sidplayfp_close()
...
+#if LIBSIDPLAYFP_CHECK_VERSION(2, 15)
+/* Reset SIDPlayFP engine
+ */
+void xs_sidplayfp_reset()
+{
+    if (state.currEng) {
+        state.currEng->reset();
+    }
+}
+#endif
```
radioactiveman: I would put the version check inside the method. You can then remove the check in xs_sidplay2.h.

```suggestion
/* Reset SIDPlayFP engine
 */
void xs_sidplayfp_reset()
{
#if LIBSIDPLAYFP_CHECK_VERSION(2, 15)
    if (state.currEng)
        state.currEng->reset();
#endif
}
```

src/sid/xmms-sid.cc
```diff
@@ -181,8 +181,33 @@ bool SIDPlugin::play(const char *filename, VFSFile &file)
...
+                    break;
+                if (bytes_skipped > bytes_to_skip)
+                    bytes_skipped = bytes_to_skip;
+                bytes_played += bytes_skipped;
+                bytes_to_skip -= bytes_skipped;
+            }              
+#else
+            AUDWARN ("Seeking requires libsidplayfp 2.15 or later, ignoring.\n");
+#endif
+        }
```
radioactiveman, line 222: The `play()` function is already long enough. Please try to extract a separate `seek()` function, e.g. like this:

```c++
class SIDPlugin : public InputPlugin
{
private:
    bool delayed_init();
    bool seek(int seek_value); // maybe also void

    Index<char> m_buffer; // replacing char *audioBuffer
    bool m_initialized = false;
    bool m_init_failed = false;
    int64_t m_bytes_played = 0;
}
```

radioactiveman (changes requested):

mvtiaine:
Updated with requested changes

radioactiveman:
> Updated with requested changes

Thank you. :+1:

@jlindgren90: Could you please review the changes as well?

Is the buffer usage correct? Should it be `.clear()` ed in `play()` like before with the `delete[]` call? Should a `Plugin::cleanup()` method be added?

radioactiveman:
@jlindgren90: You probably missed my previous review comment. Could you check it please?

jlindgren90:
I saw it but haven't had time to review, sorry.

PR #210, opened 2026-08-04
sid: Add SID model(s) as comment

mvtiaine:
also number of SIDs used (e.g. 2SID or 3SID tunes)

### DIFF
--- a/src/sid/xmms-sid.cc
+++ b/src/sid/xmms-sid.cc
@@ -225,6 +225,9 @@ static void xs_get_song_tuple_info(Tuple &tuple, const xs_tuneinfo_t &info, int 
     tuple.set_str (Tuple::Artist, info.sidComposer);
     tuple.set_str (Tuple::Copyright, info.sidCopyright);
     tuple.set_str (Tuple::Codec, info.sidFormat);
+    tuple.set_str (Tuple::Comment, info.nSIDs > 1
+        ? str_printf ("%s (%dSID)", (const char *) info.sidModel, info.nSIDs)
+        : info.sidModel);
 
     /* Get sub-tune information, if available */
     if (subTune < 0 || info.startTune > info.nsubTunes)
--- a/src/sid/xs_sidplay2.cc
+++ b/src/sid/xs_sidplay2.cc
@@ -279,6 +279,14 @@ bool xs_sidplayfp_getinfo(xs_tuneinfo_t &ti, const void *buf, int64_t bufSize)
 
     ti.sidFormat = String (myInfo->formatString());
 
+    ti.nSIDs = myInfo->sidChips();
+    if (myInfo->sidModel(0) == SidTuneInfo::SIDMODEL_6581)
+        ti.sidModel = String ("MOS6581");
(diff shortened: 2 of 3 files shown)


src/sid/xmms-sid.cc
```diff
@@ -225,6 +226,12 @@ static void xs_get_song_tuple_info(Tuple &tuple, const xs_tuneinfo_t &info, int
     tuple.set_str (Tuple::Artist, info.sidComposer);
     tuple.set_str (Tuple::Copyright, info.sidCopyright);
     tuple.set_str (Tuple::Codec, info.sidFormat);
+    if (info.nSIDs > 1) {
+        std::string comment = std::string(info.sidModel) + " (" + std::to_string(info.nSIDs) + "SID)";
+        tuple.set_str (Tuple::Comment, comment.c_str());
+    } else {
+        tuple.set_str (Tuple::Comment, info.sidModel);
+    }
```
radioactiveman: Please use `str_printf()` instead of `std::string`. And shouldn't there be a space before "SID"?

```suggestion
    tuple.set_str (Tuple::Comment, info.nSIDs > 1
        ? str_printf ("%s (%d SID)", (const char *) info.sidModel, info.nSIDs)
        : info.sidModel);
```

src/sid/xs_sidplay2.cc
```diff
@@ -281,6 +281,13 @@ bool xs_sidplayfp_getinfo(xs_tuneinfo_t &ti, const void *buf, int64_t bufSize)
 
     ti.sidFormat = String (myInfo->formatString());
 
+    ti.nSIDs = myInfo->sidChips();
+    ti.sidModel = String ("");
+    if (myInfo->sidModel(0) == SidTuneInfo::SIDMODEL_6581)
+        ti.sidModel = String ("MOS6581");
+    else if (myInfo->sidModel(0) == SidTuneInfo::SIDMODEL_8580)
+        ti.sidModel = String ("CSG8580");
+
```
radioactiveman: Set `sidModel` only once and use `String()` instead of `String("")`.

```suggestion
    if (myInfo->sidModel(0) == SidTuneInfo::SIDMODEL_6581)
        ti.sidModel = String ("MOS6581");
    else if (myInfo->sidModel(0) == SidTuneInfo::SIDMODEL_8580)
        ti.sidModel = String ("CSG8580");
    else
        ti.sidModel = String ();
```

radioactiveman (changes requested):

src/sid/xmms-sid.cc
mvtiaine: In HVSC the multi-sid tunes are named like:

```
MUSICIANS/J/Jammer/Light_Years_3SID.sid
MUSICIANS/S/Shogoon/Cowboys_Dream_2SID.sid
```

And my understanding is that 2SID or 3SID without space is kind of standard usage in C64 circles, for example: https://www.acid64.com/

```
Support for PSID, RSID, 2SID, 3SID, MUS, PRG and PC64 files
```

or https://blog.chordian.net/2017/04/01/3sid-c64-sid-music-in-9-voices/

PR #211, opened 2026-08-18
OpenPSF: Corrections for PS2 playback

Wedge009:
The full explanation for this is in https://github.com/orgs/audacious-media-player/discussions/1867.

But essentially, I found these LLM-assisted changes were necessary to resolve serious issues I found with OpenPSF:
* Missing instruments in several tracks, notably Final Fantasy X-2's 'Yuna's Theme', 'Mission Complete', 'Destruction', and Final Fantasy XI: Rise of the Zilart's 'Dash de Chocobo'.
* An overflow that resulted in a horrible screeching replacing a pipe organ instrument in Final Fantasy X-2's 'Vegnagun Starting'.
* Final Fantasy X-2's 'Crimson Squad' would hang in a way that requires Audacious to be force-killed.
* A pause after the first beat of Final Fantasy XI: Treasures of Aht Urghan's 'Run Chocobo Run!'
* A drop-silence heuristic that masked a start-up silence delay caused by idle processing, but also caused serious issues with some tracks, most notably Final Fantasy X-2's 'Chocobo'.

Unsolved:
* Handling of Final Fantasy IV's PS2 streaming driver has improved (no longer halts after a split second of playback) but still doesn't work correctly.
* Final Fantasy X cannot be decoded - the entire set remains silent (but no longer causes Audacious to hang in a way that requires a force-kill).
* The start-up silence exposed as a side-effect of removing the drop-silence heuristic is mitigated by implementation of scheduler fairness but not fully resolved.

Only the PS2 portion is affected by this, PS1 portion is untouched.

All of this may become moot if a better implementation is brought in to replace OpenPSF, as discussed in the thread linked at the start.

### DIFF
--- a/src/psf/peops2/dma.cc
+++ b/src/psf/peops2/dma.cc
@@ -104,9 +104,41 @@ EXPORT_GCC void CALLBACK SPU2writeDMA4Mem(u32 usPSXMem,int iSize)
  int i;
  u16 *ram16 = (u16 *)&psx_ram[0];
 
+ // Some streaming drivers (such as FFIV's generic PS2 streaming
+ // driver) kick a single DMA transfer covering two back-to-back 16KB
+ // per-channel halves read from one shared read-ahead buffer, but
+ // only refill that buffer's first 16KB with genuinely fresh data each
+ // cycle - the second half is whatever was last (never, for a freshly
+ // allocated buffer) written there. Past the first 8192 words (16KB) of
+ // any transfer landing in a channel's own ring, treat an exact-zero
+ // source sample as "not actually new data" and leave the existing
+ // spuMem content alone rather than overwriting real audio with a
+ // spurious zero - the correct data for that position typically arrives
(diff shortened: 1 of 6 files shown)


radioactiveman:
@kaniini: What do you think about these fixes? Should we merge them for the next 4.6.x release? Wait for 4.7? Or just focus on your [upse-ng](https://github.com/kaniini/upse-ng) plugin mentioned in [#1867](https://github.com/orgs/audacious-media-player/discussions/1867)?

Wedge009:
While OpenPSF is still the official PSF plug-in, it's horribly broken. At least these mitigate some of the problems I ran into. Replacing it with upse-ng would be a better long-term position, but I published these anyway in the hope that it will improve the OpenPSF case while it's still the official implementation. (If I knew about upse-ng I would have never attempted this as this work took a few weeks of trial and error investigation.)

On the macOS build failure - I'm not sure, but is that necessarily related to anything I've done? It looks like a download error.

PR #212, opened 2026-08-20
scrobbler2, streamtuner: honor the configured proxy settings

mikhailnov:
The scrobbler (last.fm) plugin and the Shoutcast directory model talked to the network directly (libcurl and QNetworkAccessManager respectively) and ignored the proxy configured on the Network preferences page.  Apply the use_proxy/proxy_host/proxy_port/ use_proxy_auth/proxy_user/proxy_pass/socks_proxy/socks_type settings in both places, matching what the neon and cdaudio plugins already do.

To access last.fm via proxy before this change, it was needed to do like this:

```$ https_proxy=socks5://localhost:4646 audacious```

For libcurl the settings are (re)applied before each request so that changes in the preferences take effect without restarting; when the proxy is disabled the options we set are reset, leaving libcurl's default environment-variable handling untouched.

Qt only supports SOCKS5, so the SOCKS v4a option is mapped to Socks5Proxy (most SOCKS servers accept both).

Proxy precedence after this change:

  - scrobbler2 (libcurl): the Audacious settings win when enabled (an explicitly set CURLOPT_PROXY always overrides libcurl's environment handling); when the proxy is disabled and was never enabled, libcurl keeps honoring https_proxy from the environment (unchanged behavior); toggling the proxy off in the preferences resets the options we set, which also disables env-based proxies from that point on.

  - streamtuner/shoutcast (QNetworkAccessManager): Qt never reads *_proxy environment variables, so the Audacious settings are the only source; disabled means a direct connection.

  - neon (streams, and everything going through VFS: icecast, iHeart, lyrics, ampache, album art): libneon has no environment proxy support; only the Audacious settings apply, as before.

Tested with last.fm.

Co-authored-by: Kimi AI

### DIFF
--- a/src/scrobbler2/scrobbler_communication.cc
+++ b/src/scrobbler2/scrobbler_communication.cc
@@ -29,6 +29,11 @@ typedef struct {
 
 static CURL *curlHandle = nullptr;     //global handle holding cURL options
 
+// tracks whether we set proxy options on the curl handle ourselves,
+// so that disabling the proxy in the preferences takes effect
+// without clobbering libcurl's default (environment variable) handling
+static bool proxy_applied = false;
+
 gboolean scrobbling_enabled = true;
 
 //shared variables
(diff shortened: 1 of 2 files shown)


radioactiveman:
LGTM, thanks for the contribution. :+1: Two style suggestions:

* Please extract the streamtuner changes also into a function `apply_proxy_settings ()` with an early-return if the proxy setting is disabled.

* Split the commit into two. One for scrobbler2, the other for streamtuner. And capitalize the commit message after the colon.

mikhailnov:
Done, force-pushed

PR #213, opened 2026-08-27
PipeWire plugin fixes

radioactiveman:
See the commit messages for more details.

CC'ing @maris-ab as author of the current drain() implementation.


src/pipewire/pipewire.cc
```diff
@@ -205,28 +206,24 @@ void PipeWireOutput::drain()
...
-    while ((buflen = m_buffer.len()) > 0)
-    {
-        pw_thread_loop_timed_wait(m_loop, 1);
-        if (buflen <= m_buffer.len())
-        {
-            AUDERR("PipeWireOutput: buffer drain lock\n");
-            break;
-        }
-    }
+    while (m_buffer.len() > 0)
```
jlindgren90, line 209: Is there any risk of any infinite loop/hang here, if the stream stalls/errors?

src/pipewire/pipewire.cc
```diff
@@ -205,28 +206,24 @@ void PipeWireOutput::drain()
...
+        pw_thread_loop_wait(m_loop);
 
+    m_drained = false;
     pw_stream_flush(m_stream, true);
-    pw_thread_loop_timed_wait(m_loop, 1); // trigger on_drained() callback
+
+    while (!m_drained)
```
jlindgren90, line 215: Same concern here, can this loop forever?

jlindgren90 (commented):
I'm still not very familiar with the PipeWire API, so just had a couple of questions. If you're happy with it, go ahead and merge.

src/pipewire/pipewire.cc
kaniini: yeah these scare me

